PR #133, opened 2022-12-14
feat: Forex goes back when looking for rates, reports the actual timestamp for rates

yotamhc:
This PR adds the following:
1. When looking up a forex rate, and no rates exist for the requested date, it will go back up to 4 days as the requested date may fall on a weekend or a holiday.
2. The effective timestamp for which forex rates are reported are added to the `QueriedExchangeRate` struct and consequently later to the `ExchangeRateMetadata` struct, so the user can see what date the forex rates that were used are from.

### DIFF
--- a/src/xrc/src/forex.rs
+++ b/src/xrc/src/forex.rs
@@ -304,34 +304,38 @@ impl ForexRateStore {
                 let quote = rates_for_timestamp.get(&quote_asset);
 
                 match (base, quote) {
-                    (Some(base_rate), Some(quote_rate)) => return Ok(base_rate.clone() / quote_rate.clone()),
+                    (Some(base_rate), Some(quote_rate)) => {
+                        return Ok(base_rate.clone() / quote_rate.clone())
+                    }
                     (Some(base_rate), None) => {
                         // If the quote asset is USD, it should not be present in the map and the base rate already uses USD as the quote asset.
                         if quote_asset == USD {
-                            return Ok(base_rate.clone())
+                            return Ok(base_rate.clone());
                         } else {
                             return Err(GetForexRateError::CouldNotFindQuoteAsset(
                                 timestamp,
                                 quote_asset.to_string(),
-                            ))
+                            ));
                         }
                     }
-                    (None, Some(_)) => return Err(GetForexRateError::CouldNotFindBaseAsset(
-                        timestamp,
-                        base_asset.to_string(),
-                    )),
+                    (None, Some(_)) => {
+                        return Err(GetForexRateError::CouldNotFindBaseAsset(
+                            timestamp,
+                            base_asset.to_string(),
+                        ))
+                    }
                     (None, None) => {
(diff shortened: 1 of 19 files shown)


src/xrc/src/forex.rs
```diff
@@ -288,48 +289,63 @@ impl ForexRateStore {
...
+                                timestamp,
+                                base_asset.to_string(),
+                                quote_asset.to_string(),
+                            ));
+                        }
+                    }
                 }
+            } else {
+                // No forex rates were found for the query timestamp.
+                continue;
```
dfinity-ryancroote, line 344: Is the continue necessary? It seems pretty clear that no rates were found and the loop is going to continue.

yotamhc: Yes you are right, it is not necessary at all.

src/xrc/src/forex.rs
```diff
@@ -1861,14 +1883,63 @@ mod test {
...
+                    forex_timestamp: Some(1234),
                 }
             },
         );
 
         assert_eq!(store.allocated_bytes(), 273);
     }
 
-    /// This functiont ests the the forexes can report the max response bytes needed
+    /// This function tests the "go back" mechanism where when there are not rates for a requested timestamp, we may go back up to [MAX_DAYS_TO_GO_BACK] days.
```
THLO, line 1894:
```suggestion
    /// This function tests the "go back" mechanism where, when there are no rates for a requested timestamp, we may go back up to [MAX_DAYS_TO_GO_BACK] days.
```

src/xrc/src/forex.rs
```diff
@@ -1861,14 +1883,63 @@ mod test {
...
+        assert_eq!(
+            store
+                .get(queried_timestamp, queried_timestamp, "EUR", USD)
+                .unwrap()
+                .forex_timestamp
+                .unwrap(),
+            timestamp
+        );
+        // But also that we cannot retrieve rates for more than that.
+        let queried_timestamp = queried_timestamp + SECONDS_PER_DAY;
```
THLO, line 1935: If it's one second less, it would still find the rate because the timestamps are "rounded" to the start of the day, right?

src/xrc/src/forex.rs
```diff
@@ -1861,14 +1883,63 @@ mod test {
...
+        assert!(matches!(
+            store.get(queried_timestamp, queried_timestamp, "EUR", USD),
+            Err(GetForexRateError::InvalidTimestamp(_queried_timestamp))
+        ));
+    }
+
+    /// This function tests the the forexes can report the max response bytes needed
```
THLO, line 1942:
```suggestion
    /// This function tests that the forex sources can report the max response bytes needed.
```

THLO (approved):

src/xrc/src/forex.rs
yotamhc: Yes

PR #134, opened 2022-12-14
refactor: Rate limiting guard should track the calculated HTTP requests and add `is_available` to exchanges and forex sources to clean up feature flag check.

dfinity-ryancroote:
This PR refactors the following two items:

1. The rate limiting guard now takes the number of HTTP requests needed instead of the number of rates needed. This moves the calculation out. This also fixes a bug when IPv4 support is not available causing more requests to be reserved than required.

2. It adds an `is_available` to `Exchange` and `Forex` to make it easier to determine if the exchange or forex can be used as the IPv4 feature flag was used to compile the canister or they support IPv6 out-of-the-box.

### DIFF
--- a/src/xrc/src/api.rs
+++ b/src/xrc/src/api.rs
@@ -269,8 +269,10 @@ async fn handle_cryptocurrency_pair(
         num_rates_needed = num_rates_needed.saturating_add(1);
     }
 
+    let available_exchanges_count = EXCHANGES.iter().filter(|e| e.is_available()).count();
+    let http_requests_needed = num_rates_needed.saturating_mul(available_exchanges_count);
     if !utils::is_caller_the_cmc(&caller) {
-        if is_rate_limited(num_rates_needed) {
+        if is_rate_limited(http_requests_needed) {
             return Err(ExchangeRateError::RateLimited);
         }
         env.charge_cycles(num_rates_needed)?;
@@ -284,8 +286,6 @@ async fn handle_cryptocurrency_pair(
 
     let base_asset = base_asset.clone();
     let quote_asset = quote_asset.clone();
-    let available_exchanges_count = EXCHANGES.iter().filter(|e| e.is_available()).count();
-    let http_requests_needed = num_rates_needed.saturating_mul(available_exchanges_count);
     with_request_counter(http_requests_needed, async move {
         let base_rate = match maybe_base_rate {
             Some(base_rate) => base_rate,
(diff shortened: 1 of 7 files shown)


src/xrc/src/api.rs
```diff
@@ -370,7 +374,9 @@ async fn handle_crypto_base_fiat_quote_pair(
     }
 
     let base_asset = base_asset.clone();
-    with_request_counter(num_rates_needed, async move {
+    let available_exchanges_count = EXCHANGES.iter().filter(|e| e.is_available()).count();
+    let http_requests_needed = num_rates_needed.saturating_mul(available_exchanges_count);
```
THLO, line 385: So now we have these two lines three times outside of the guard as opposed to twice inside the guard (in the functions `new` and `drop`). Is there still a net gain...?

src/xrc/src/exchanges.rs
```diff
@@ -114,6 +114,12 @@ macro_rules! exchanges {
                     $(Exchange::$name(exchange) => exchange.max_response_bytes()),*,
                 }
             }
+
+            /// This method returns whether the exchange should be called.
+            /// Currently, only IPv6 support determines whether is should be used.
```
THLO, line 119: This comment is misleading because there is a check if IPv4 is available. It just happens that it always return `false` right now.
I think it's better to put such a comment together with the `is_ipv4_support_available` function so that the comment is updated together with the implementation.

src/xrc/src/forex.rs
```diff
@@ -180,6 +182,12 @@ macro_rules! forex {
                     $(Forex::$name(forex) => forex.max_response_bytes()),*,
                 }
             }
+
+            /// This method returns whether the exchange should be called.
+            /// Currently, only IPv6 support determines whether is should be used.
```
THLO, line 187: Same here.

THLO (approved):

src/xrc/src/api.rs
dfinity-ryancroote: I do not think there is a gain. I see it more as a loss at this point (up to the programmer to supply the request amount). I am going to move it back into the guard as that seems safer long term.

src/xrc/src/exchanges.rs
dfinity-ryancroote: Makes sense. I have updated the comments.

THLO (approved):

nategurian (approved):

PR #136, opened 2022-12-15
refactor: Clean up errors from transforms

dfinity-ryancroote:
This PR removes an unused arm from `CallExchangeError` and provides an error to read format when calling an exchange fails.


THLO (approved):
Nice!

PR #138, opened 2022-12-16
feat: Add `canister_inspect_message`

dfinity-ryancroote:
This PR uses `canister_inspect_message` to ensure only messages from other canisters or requests to the metrics are allowed.

### DIFF
--- a/src/xrc/src/main.rs
+++ b/src/xrc/src/main.rs
@@ -40,6 +40,13 @@ pub fn http_request(request: xrc::types::HttpRequest) -> xrc::types::HttpRespons
     xrc::http_request(request)
 }
 
+/// Inspect ingress messages coming in to ensure that only messages from other canisters or requests
+/// to the metrics are allowed.
+///
+/// https://internetcomputer.org/docs/current/references/ic-interface-spec/#system-api-inspect-message
+#[ic_cdk_macros::inspect_message]
+pub fn inspect_message() {}
+
 fn main() {}
 
 #[cfg(test)]


nategurian (approved):

src/xrc/src/main.rs
```diff
@@ -40,6 +40,13 @@ pub fn http_request(request: xrc::types::HttpRequest) -> xrc::types::HttpRespons
     xrc::http_request(request)
 }
 
+/// Inspect ingress messages coming in to ensure that only messages from other canisters or requests
+/// to the metrics are allowed.
+///
+/// https://internetcomputer.org/docs/current/references/ic-interface-spec/#system-api-inspect-message
+#[ic_cdk_macros::inspect_message]
+pub fn inspect_message() {}
```
THLO, line 48: Doesn't this implementation also drop all ingress messages to the metrics endpoint?

THLO: I think we need a test to ensure that metrics are accessible.

dfinity-ryancroote: The metrics will still be available. From the interface spec: `The canister_inspect_message is not invoked for query calls, inter-canister calls or calls to the management canister.`. The metrics are served via a query call. `http_request` uses query calls only UNLESS the canister upgrades the call to an update (this can be done through `icx-proxy`).

dfinity-ryancroote: We can still test it if you'd like of course. We would simply make a query call to the `http_request` endpoint or curl `http://127.0.0.1:8000/metrics?canisterId=ryjl3-tyaaa-aaaaa-aaaba-cai`.

THLO (approved):

PR #142, opened 2023-01-04
refactor: Move unit test xml & json to separate files.

dfinity-ryancroote:
This PR moves the rather large string blobs from the unit tests into separate files.


THLO (approved):
A simple but beautiful change.
Thanks!

PR #144, opened 2023-01-05
Switching to an LRU cache

THLO:
This PR replaces the current time-based, custom cache with a standard LRU cache to retain exchange rates longer.

The cache will be made persistent across updates in a follow-up PR.


dfinity-ryancroote (approved):
I like all of the red. 😄

PR #145, opened 2023-01-05
feat(refactor): Clean up fee calculation and apply a minimum fee for errors.

dfinity-ryancroote:
This PR cleans up the fee calculation in case there is an error when pulling a rate from the `ForexRateStore`. It cleans up the fee calculation by adding a `ChargeOption` enum that maps more easily to how the fee is calculated instead of having multiple arguments.


src/xrc/src/api.rs
```diff
@@ -342,12 +346,19 @@ async fn handle_crypto_base_fiat_quote_pair(
...
+    let forex_rate = match forex_rate_result {
+        Ok(forex_rate) => forex_rate,
+        Err(_) => {
+            env.charge_cycles(ChargeOption::MinimumFee)?;
+            return forex_rate_result;
+        }
+    };
 
+    let maybe_crypto_base_rate =
+        with_cache_mut(|cache| cache.get(&base_asset.symbol, timestamp, time));
```
THLO, line 361:
```suggestion
        with_cache_mut(|cache| cache.get(&(base_asset.symbol.clone(), timestamp)).cloned());
```

src/xrc/src/environment.rs
```diff
@@ -69,34 +65,39 @@ pub(crate) trait Environment {
     }
 }
 
+/// Used to determine what should be charged when calculating the fee.
```
THLO, line 68: Some more explanation about the different fields would be nice here.

src/xrc/src/lib.rs
```diff
@@ -63,8 +63,8 @@ pub const XRC_IMMEDIATE_REFUND_CYCLES: u64 = 5_000_000_000;
 /// The base cost in cycles that will always be charged when using the `xrc` canister.
```
THLO, line 63: This comment is not really correct unless "using" means "getting a valid response".

src/xrc/src/environment.rs
```diff
@@ -69,34 +65,39 @@ pub(crate) trait Environment {
...
+pub(crate) enum ChargeOption {
+    MinimumFee,
+    BaseCost,
+    OutboundRatesNeeded(usize),
+}
+
 /// This function calculates the fee based off the number of outbound requests needed in order
```
THLO, line 75:
```suggestion
/// This function calculates the fee based on the number of outbound requests needed in order
```

THLO (approved):

src/xrc/src/lib.rs
dfinity-ryancroote: Updated the comment.

PR #146, opened 2023-01-06
feat: Add forex rate pair unit tests

dfinity-ryancroote:
This PR adds unit tests for forex rate pairs when calling `get_exchange_rate`.

### DIFF
--- a/src/xrc/src/api/test.rs
+++ b/src/xrc/src/api/test.rs
@@ -711,7 +711,7 @@ fn get_exchange_rate_for_fiat_with_unknown_symbol() {
 }
 
 /// This function tests to ensure the minimum fee cost is accepted and an error is returned when
-/// an unknown timestamp.
+/// a timestamp is not known to the forex store.
 #[test]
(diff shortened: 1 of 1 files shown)


src/xrc/src/api/test.rs
```diff
@@ -597,3 +597,171 @@ fn get_crypto_fiat_pair_fails_when_the_fiat_timestamp_is_not_known() {
...
+        .expect("future should complete");
+    assert!(
+        matches!(result, Err(ExchangeRateError::ForexBaseAssetNotFound)),
+        "Received the following result: {:#?}",
+        result
+    );
+}
+
+/// This function tests to ensure the minimum fee cost is accepted and an error is returned when
+/// an unknown timestamp.
```
THLO, line 714: an unknown timestamp...?

THLO (approved):

src/xrc/src/api/test.rs
dfinity-ryancroote: Fixed the comment. It should be `a timestamp that is not known to the forex store`.

PR #152, opened 2023-01-17
feat: Add the ability to mark requests as pending

dfinity-ryancroote:
This PR adds the ability to track pending requests. A request is determined as `pending` when one of the assets is currently in process of being retrieved from the exchanges.

For example, the XRC receives a request `X` for `ICP/BTC` at timestamp `100`. While the XRC starts processing request `X`, another request (`Y`) for `BTC/USDT` at timestamp `100` is received. The XRC canister adds the following symbol-timestamp pairs to the tracking set from request `X`: 
`("ICP", 100)` and `("BTC", 100)`. 

While waiting for request `X`'s results, the XRC begins to process request `Y`. As request `Y` contains the symbol-timestamp pair `("BTC", 100)`, request `Y` is determined to be `pending` and the `Pending` error is returned to the user.

### DIFF
--- a/src/xrc/src/inflight.rs
+++ b/src/xrc/src/inflight.rs
@@ -1,0 +1,62 @@
+use std::{
+    cell::RefCell,
+    collections::{HashMap, HashSet},
+};
+
+use crate::{candid::ExchangeRateError, QueriedExchangeRate};
+
+type InflightCryptoUsdtRequests = HashSet<(String, u64)>;
+
+thread_local! {
+    static INFLIGHT_CRYPTO_USDT_RATE_REQUESTS: RefCell<InflightCryptoUsdtRequests> = RefCell::new(HashSet::new());
+}
+
+fn contains(key: &(String, u64)) -> bool {
+    INFLIGHT_CRYPTO_USDT_RATE_REQUESTS.with(|cell| cell.borrow().contains(key))
+}
+
+fn add(key: (String, u64)) {
+    INFLIGHT_CRYPTO_USDT_RATE_REQUESTS.with(|cell| {
+        cell.borrow_mut().insert(key);
+    });
+}
+
+fn remove(key: &(String, u64)) {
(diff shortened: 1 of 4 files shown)


src/xrc/src/inflight.rs
```diff
@@ -0,0 +1,152 @@
...
+        for symbol in &symbols {
+            add((symbol.clone(), timestamp));
+        }
+        Self { symbols, timestamp }
+    }
+}
+
+impl Drop for InflightCryptoUsdtRequestsGuard {
+    /// Removes all symbols paired to a given timestamp to the tracking set.
+    fn drop(&mut self) {
```
THLO, line 78: Does the guard pattern completely alleviate Dimitris' concerns?
Is the guard guaranteed to be removed even if there is a panic after an asynchronous call?

src/xrc/src/inflight.rs
```diff
@@ -0,0 +1,152 @@
...
+        assert!(!contains(&("ICP".to_string(), 0)));
+        assert!(!contains(&("BTC".to_string(), 0)));
+    }
+
+    /// The function verifies that when an error is returned from the provided async block,
+    /// the guard correctly releases the symbol-timestamp pair from the set.
+    #[test]
+    fn with_inflight_tracking_with_error_result_returned() {
+        let err =
+            with_inflight_tracking(vec!["ICP".to_string(), "BTC".to_string()], 0, async move {
```
THLO, line 117: What if a function that requires `await` is called in this block and there is a `panic` after the `await`?

THLO (approved):
Very nice PR description!

src/xrc/src/inflight.rs
dfinity-ryancroote: I have added a new test that covers this concern. The guard does guarantee that the pairs are removed after an async call. When the panic occurs, rust will unwind. The unwind will cause the `drop` to be called.

PR #158, opened 2023-01-23
feat: Add crypto/USDT unit test

dfinity-ryancroote:
This PR adds a unit test when a `crypto/USDT` pair is requested. The intended functionality is that the canister will only call for the `crypto/USDT` from the exchanges as `USDT` is the base unit.

### DIFF
--- a/src/xrc/src/api/test.rs
+++ b/src/xrc/src/api/test.rs
@@ -867,3 +867,42 @@ fn get_exchange_rate_will_retrieve_rates_if_inflight_tracking_contains_any_symbo
         .expect("future should complete");
     assert!(matches!(result, Err(ExchangeRateError::Pending)));
 }
+
+/// This function tests that [get_exchange_rate] can retrieve */USDT rates with one set of outbound
+/// calls.
+#[test]
+fn get_exchange_rate_can_retrieve_icp_usdt() {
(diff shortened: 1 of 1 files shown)


src/xrc/src/api/test.rs
```diff
@@ -867,3 +867,42 @@ fn get_exchange_rate_will_retrieve_rates_if_inflight_tracking_contains_any_symbo
...
+    let call_exchanges_impl = TestCallExchangesImpl::builder()
+        .with_get_cryptocurrency_usdt_rate_responses(hashmap! {
+            "ICP".to_string() => Ok(icp_queried_exchange_rate_mock())
+        })
+        .build();
+    let env = TestEnvironment::builder()
+        .with_cycles_available(XRC_REQUEST_CYCLES_COST)
+        .with_accepted_cycles(XRC_BASE_CYCLES_COST + XRC_OUTBOUND_HTTP_CALL_CYCLES_COST)
+        .build();
+    let request = GetExchangeRateRequest {
```
THLO, line 884: What if the base and quote assets are reversed?
The same behavior should be observed in this case.

THLO (approved):

src/xrc/src/api/test.rs
dfinity-ryancroote: Yes. Added a test to show this is the case.

PR #159, opened 2023-01-24
Use BTreeMap for forex rate serialization so that they are sorted by key

yotamhc:


dfinity-ryancroote:
I think we can close this down. The other forexes do not experience this issue.

PR #163, opened 2023-01-26
feat: Removing logging for `MonetaryAuthorityOfSingapore`

dfinity-ryancroote:
This PR removes logging for communicating with the forex `MonetaryAuthorityOfSingapore`. The logging is no longer needed as we have determined that the forex is syncing correctly.


THLO:
Could you please add an explanation why logging is removed?

THLO (approved):

PR #166, opened 2023-01-30
feat: Allow e2e-tests to take arguments so one test can be ran easily

dfinity-ryancroote:
This PR adds the ability to `./scripts/e2e-tests` to run a singular test. During development of new `e2e` tests, it is generally painful to run a singular test. This is addressed by giving the option to add a flag (`n`) to provide a name of the test that a user wants to run.

Example: `./scripts/e2e-tests -n test_name_here`

### DIFF
--- a/scripts/e2e-tests
+++ b/scripts/e2e-tests
@@ -5,6 +5,22 @@
 # Builds the wasm locally, the e2e base image, moves the wasm to the
 # appropriate location, and runs the tests
 
+name=''
+
+help() {
+  echo "Usage: ..."
+}
+
+while getopts 'hn:' flag; do
+  case "${flag}" in
+    n) name="${OPTARG}" ;;
+    h) help
+       exit 1 ;;
+    *) help
+       exit 1 ;;
+  esac
+done
+
 # Build the wasm without needing a canister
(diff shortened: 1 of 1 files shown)


scripts/e2e-tests
```diff
@@ -5,6 +5,24 @@
 # Builds the wasm locally, the e2e base image, moves the wasm to the
 # appropriate location, and runs the tests
 
+name=''
+
+help() {
+  echo "Usage:"
+  echo "    To run all tests:       ./scripts/e2e-tests"
+  echo "    To run a singular test: ./scripts/e2e-tests -n test_name_here"
```
THLO, line 13:
```suggestion
  echo "    To run a singular test: ./scripts/e2e-tests -n test_name"
```

THLO: There are different notations for parameters (TEST_NAME, [test name],  ...). I don't have a strong preference. We should simply use consistent notation.

THLO (approved):

scripts/e2e-tests
dfinity-ryancroote: Decided to go with `TEST_NAME`.

PR #169, opened 2023-01-31
refactor: Allow e2e tests to have different response bodies beyond JSON.

dfinity-ryancroote:
This PR updates the e2e tests response builder to allow for different content types when responding to the canister HTTP outcalls. The PR was created to start adding forex sources (some sources return XML) to the e2e tests.


THLO (approved):
Nice!

PR #174, opened 2023-02-17
`feature`: If timestamp = null and the request is inflight, use the previous minute in hopes to return a rate.

dfinity-ryancroote:
This PR implements the following:

1. If the request contains a timestamp = null, the XRC will get the current normalized timestamp. 
2. If the requested assets with the current normalized timestamp are considered to be inflight, the XRC then goes back 1 additional minute (labeled internally as a past normalized timestamp).
3. If requested assets with the past normalized timestamp *are not* in the cache, the minimum fee is charged and the caller receives a pending error.
4. If requested assets with the past normalized timestamp *are* in the cache, the base cost fee is charged and the caller receives the requested rate.

### DIFF
--- a/src/xrc/src/api.rs
+++ b/src/xrc/src/api.rs
@@ -235,12 +235,13 @@ fn validate(rate: QueriedExchangeRate) -> Result<QueriedExchangeRate, ExchangeRa
     }
 }
 
-#[derive(PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq)]
 enum TimestampType {
     Current,
     Previous,
 }
 
+#[derive(Debug)]
 struct Timestamp {
     value: u64,
     r#type: TimestampType,
@@ -304,10 +305,15 @@ async fn handle_cryptocurrency_pair(
     }
 
     if !utils::is_caller_privileged(&caller) {
+        println!("{:#?}", timestamp);
         let rate_limited = is_rate_limited(num_rates_needed);
         let already_inflight = is_request_already_inflight(request, timestamp.value);
         let is_previous_minute_with_rates_needed =
             timestamp.r#type == TimestampType::Previous && num_rates_needed > 0;
+        println!(
+            "is_previous_minutes_with_rates_needed: {}",
+            is_previous_minute_with_rates_needed
+        );
         let charge_cycles_option =
             if rate_limited || already_inflight || is_previous_minute_with_rates_needed {
                 ChargeOption::MinimumFee
--- a/src/xrc/src/api.rs
+++ b/src/xrc/src/api.rs
@@ -273,7 +273,7 @@ impl NormalizedTimestamp {
 /// If the request contains a timestamp, the function returns the normalized requested timestamp.
 /// If the request's timestamp is null, the function returns the current timestamp if no assets are found
 /// to be inflight; otherwise, the normalized timestamp from one minute ago.
-fn get_normalized_timestamp_for_crypto_assets(
+fn get_normalized_timestamp(
     env: &impl Environment,
     request: &GetExchangeRateRequest,
 ) -> NormalizedTimestamp {
@@ -294,7 +294,7 @@ async fn handle_cryptocurrency_pair(
     call_exchanges_impl: &impl CallExchanges,
     request: &GetExchangeRateRequest,
 ) -> Result<QueriedExchangeRate, ExchangeRateError> {
-    let timestamp = get_normalized_timestamp_for_crypto_assets(env, request);
+    let timestamp = get_normalized_timestamp(env, request);
 
     let caller = env.caller();
     let (maybe_base_rate, maybe_quote_rate) = with_cache_mut(|cache| {
@@ -387,7 +387,7 @@ async fn handle_crypto_base_fiat_quote_pair(
     call_exchanges_impl: &impl CallExchanges,
     request: &GetExchangeRateRequest,
 ) -> Result<QueriedExchangeRate, ExchangeRateError> {
-    let timestamp = get_normalized_timestamp_for_crypto_assets(env, request);
+    let timestamp = get_normalized_timestamp(env, request);
     let caller = env.caller();
 
     let forex_rate_result = with_forex_rate_store(|store| {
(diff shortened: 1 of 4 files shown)


src/xrc/src/api.rs
```diff
@@ -235,17 +236,76 @@ fn validate(rate: QueriedExchangeRate) -> Result<QueriedExchangeRate, ExchangeRa
     }
 }
 
+#[derive(Debug, PartialEq, Eq)]
+enum NormalizedTimestampType {
+    /// The timestamp is within the past minute.
+    RequestedOrCurrent,
+    /// The timestamp is within 1-2 minutes ago.
```
THLO, line 243: Why does `Past` mean "the timestamp is within 1-2 minutes ago"?

src/xrc/src/api.rs
```diff
@@ -235,17 +236,76 @@ fn validate(rate: QueriedExchangeRate) -> Result<QueriedExchangeRate, ExchangeRa
...
+    Past,
+}
+
+#[derive(Debug)]
+struct NormalizedTimestamp {
+    /// The timestamp in seconds.
+    value: u64,
+    /// Used to determine if the timestamp is a current timestamp or a timestamp from a minute ago.
```
THLO, line 251: Based on this statement, you need a `Current` and a `Recent` type.

src/xrc/src/api.rs
```diff
@@ -235,17 +236,76 @@ fn validate(rate: QueriedExchangeRate) -> Result<QueriedExchangeRate, ExchangeRa
...
+        Self {
+            value,
+            r#type: NormalizedTimestampType::Past,
+        }
+    }
+}
+
+/// If the request contains a timestamp, the function returns the normalized requested timestamp.
+/// If the request's timestamp is null, the function returns the current timestamp if no assets are found
+/// to be inflight, OR the normalized timestamp from one minute ago.
```
THLO, line 275: "OR" is a bit confusing. It sounds like the function is allowed to return either.
How about "otherwise"?

src/xrc/src/api.rs
THLO: Having read the code below, "within 1-2 minutes ago" refers to how we actually use this type.
I think the comment is confusing here. It should simply say that it refers to a timestamp in the past.

src/xrc/src/api.rs
```diff
@@ -235,17 +236,76 @@ fn validate(rate: QueriedExchangeRate) -> Result<QueriedExchangeRate, ExchangeRa
...
+    env: &impl Environment,
+    request: &GetExchangeRateRequest,
+) -> NormalizedTimestamp {
+    if request.timestamp.is_some() {
+        let timestamp = utils::get_normalized_timestamp(env, request);
+        return NormalizedTimestamp::requested_or_current(timestamp);
+    }
+
+    let timestamp = utils::get_normalized_timestamp(env, request);
+    let base_asset_is_inflight = matches!(request.base_asset.class, AssetClass::Cryptocurrency)
```
THLO, line 286: Why do you use `matches!` here?
Why is this check even needed? Only cryptocurrencies can be in-flight, right? So, if the second check returns `true`, the first must be `true`.

src/xrc/src/api.rs
```diff
@@ -260,19 +320,23 @@ async fn handle_cryptocurrency_pair(
...
+                ChargeOption::OutboundRatesNeeded(num_rates_needed)
+            };
+
         env.charge_cycles(charge_cycles_option)?;
         if rate_limited {
             return Err(ExchangeRateError::RateLimited);
         }
 
-        if already_inflight {
+        if already_inflight || is_past_timestamp_with_rates_needed {
```
THLO, line 339: How about `is_past_timestamp_not_cached` instead ("cached" implies that no rates are needed)?
It would be clearer here that we return `Pending` if the past time slot cannot be served from the cache.

src/xrc/src/api.rs
```diff
@@ -373,18 +437,22 @@ async fn handle_crypto_base_fiat_quote_pair(
...
+                ChargeOption::OutboundRatesNeeded(num_rates_needed)
+            };
+
         env.charge_cycles(charge_cycles_option)?;
         if rate_limited {
             return Err(ExchangeRateError::RateLimited);
         }
 
-        if already_inflight {
+        if already_inflight || is_past_minute_with_rates_needed {
```
THLO, line 455: Same here.

src/xrc/src/api/test.rs
```diff
@@ -945,3 +945,260 @@ fn get_exchange_rate_can_retrieve_usdt_icp() {
...
+                symbol: "ICP".to_string(),
+                class: AssetClass::Cryptocurrency,
+            },
+            timestamp: None,
+        };
+
+        let result = get_exchange_rate_internal(&env, &call_exchanges_impl, &request)
+            .now_or_never()
+            .expect("future should complete");
+        assert!(matches!(result, Ok(_)));
```
THLO, line 1062: How about checking the timestamp in the result? It should be the correct timestamp associated with the returned rate.

src/xrc/src/lib.rs
```diff
@@ -98,6 +98,9 @@ const RATE_UNIT: u64 = 10u64.saturating_pow(DECIMALS);
 /// Used for setting the max response bytes for the exchanges and forexes.
 const ONE_KIB: u64 = 1_024;
 
+// 1 minute in seconds
+const ONE_MINUTE: u64 = 60;
```
THLO, line 102: Just in case we don't remember how many seconds there are in a minute.   :-)

THLO (approved):

src/xrc/src/api.rs
dfinity-ryancroote: Makes sense. Updated.

src/xrc/src/api.rs
dfinity-ryancroote: Good point. Removed.

src/xrc/src/api/test.rs
dfinity-ryancroote: Added checks. 👍

src/xrc/src/api.rs
dfinity-ryancroote: That makes sense. Updated.

PR #175, opened 2023-02-21
`bugfix`: fiat-crypto pairs were not handled correctly in earlier refactor

dfinity-ryancroote:
When deploying the timestamp = null feature, ForexQuoteAssetNotFound errors started to populate the logs. This lead to the discovery that the refactor missed the inversion of the base and quote assets for fiat-crypto pairs.

This PR addresses the issue by inverting the base and quote assets for a fiat-crypto pair. The inverted request is then passed in to the `handle_crypto_base_fiat_quote_pair` handle function. To ensure coverage to avoid this problem from occurring in the future, additional unit tests were added to ensure this case is covered.

### DIFF
--- a/src/xrc/src/api.rs
+++ b/src/xrc/src/api.rs
@@ -193,6 +193,16 @@ async fn get_exchange_rate_internal(
     result.map(|r| r.into())
 }
 
+/// This function is used for handling fiat-crypto pairs.
+fn invert_assets_in_request(request: &GetExchangeRateRequest) -> GetExchangeRateRequest {
+    GetExchangeRateRequest {
+        base_asset: request.quote_asset.clone(),
+        quote_asset: request.base_asset.clone(),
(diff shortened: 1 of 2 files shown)


nategurian (approved):

src/xrc/src/api.rs
```diff
@@ -213,12 +223,13 @@ async fn route_request(
                 })
         }
         (AssetClass::FiatCurrency, AssetClass::Cryptocurrency) => {
-            handle_crypto_base_fiat_quote_pair(env, call_exchanges_impl, request)
+            let inverted_request = invert_assets_in_request(request);
+            handle_crypto_base_fiat_quote_pair(env, call_exchanges_impl, &inverted_request)
                 .await
                 .map(|r| r.inverted())
```
THLO, line 229: Nit: A more expressive variable name than `r` would be preferable.

PR #177, opened 2023-02-23
`feat`: Add `disable-forex-timezone-offset` for system testing purposes

dfinity-ryancroote:
This PR adds a feature flag to disable the forex timezone offset. This causes issues in determining what values will be used when creating system tests that expect hand built responses.

### DIFF
--- a/src/xrc/src/forex.rs
+++ b/src/xrc/src/forex.rs
@@ -165,6 +165,9 @@ macro_rules! forex {
 
             /// This method invokes the forex's [IsForex::offset_timestamp_to_timezone] function.
             pub fn offset_timestamp_to_timezone(&self, timestamp: u64) -> u64 {
+                if cfg!(feature = "disable-forex-weekend-check") {
+                    return timestamp;
+                }
                 match self {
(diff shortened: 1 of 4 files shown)


src/xrc/src/forex.rs
```diff
@@ -275,11 +281,13 @@ impl ForexRateStore {
 
         // If today's date is requested, and the day is not over anywhere on Earth, use yesterday's date
         // Get the normalized timestamp for yesterday.
-        let yesterday = (current_timestamp as i64 + TIMEZONE_AOE_SHIFT_SECONDS) as u64
-            / SECONDS_PER_DAY
-            * SECONDS_PER_DAY;
-        if timestamp > SECONDS_PER_DAY && yesterday == timestamp {
-            timestamp -= SECONDS_PER_DAY;
+        if !cfg!(feature = "disable-forex-timezone-offset") {
```
THLO, line 284: The comment only applies to the code executed when the forex timezone offset is enabled.

THLO (approved):

src/xrc/src/forex.rs
dfinity-ryancroote: Fixed.

PR #179, opened 2023-02-28
Require a minimum number of sources for privileged canisters when serving from the cache

THLO:
There is a risk that an attacker might try to get an exchange rate based on the rates of a small number of exchanges into the cache by timing the request carefully.
The CMC might then be served this rate.

This PR changes adds the restriction that cached rates may only be served to all privileged canisters if they are based on a specific minimum number of collected rates.

### DIFF
--- a/src/xrc/src/api.rs
+++ b/src/xrc/src/api.rs
@@ -20,6 +20,9 @@ use futures::future::join_all;
 /// The expected base rates for stablecoins.
 const STABLECOIN_BASES: &[&str] = &[DAI, USDC];
 
+/// A cached rate is only used for privileged canisters if there are at least this many source rates.
+const MIN_NUM_RATES_FOR_PRIVILEGED_CANISTERS: usize = 3;
+
 #[async_trait]
 trait CallExchanges {
     async fn get_cryptocurrency_usdt_rate(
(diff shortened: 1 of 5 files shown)


src/xrc/src/api.rs
```diff
@@ -300,6 +303,25 @@ fn get_normalized_timestamp(
...
+        {
+            Some(rate.to_owned())
+        } else {
+            None
+        }
+    } else {
+        None
+    }
+}
+
```
dfinity-ryancroote, line 324: How about something like this instead?

```rust
fn get_rate_from_cache(
    cache: &mut ExchangeRateCache,
    caller: &Principal,
    symbol: &str,
    timestamp: u64,
) -> Option<QueriedExchangeRate> {
    let maybe_rate = cache.get(symbol, timestamp);
    match maybe_rate {
        Some(ref rate) => {
            if !utils::is_caller_privileged(caller) {
                return maybe_rate;
            }

            if rate.base_asset.symbol == USDT {
                return maybe_rate;
            }

            if rate.rates.len() >= MIN_NUM_RATES_FOR_PRIVILEGED_CANISTERS {
                Some(rate.to_owned())
            } else {
                None
            }
        }
        None => None,
    }
}
```

Then inside the `with_cache_mut` block you would then just have:

```rust
        let maybe_base_rate =
            get_rate_from_cache(cache, &caller, &request.base_asset.symbol, timestamp.value);
```

dfinity-ryancroote (approved):

PR #180, opened 2023-02-28
Adding the Poloniex exchange

THLO:
This PR adds support for the Poloniex exchange.

While the trading volume is low, it appears that the exchange even returns a reasonable exchange rate if there was no trade in the queried time interval.


dfinity-ryancroote (approved):
Looks good. Obviously tests need to be fixed. 😄

THLO:
> Looks good. Obviously tests need to be fixed. smile

Yes, adding an exchange always breaks the tests...
I'll look into it tomorrow.

PR #182, opened 2023-03-09
bugfix: Address minor bugs when querying ICP/XDR rates

dfinity-ryancroote:
This PR addresses the following behaviors we have seen when testing the ICP/XDR rates for the CMC:

* `base_asset_num_received_rates` should be equal to `quote_asset_num_received_rates`
* The timezone adjustment was shifting the timestamp by 12 days into the past instead of 12 hours. This would cause the requested timestamp to always fall into the most recent forex rates instead of using the prior day's rates.
* The forex rate store was accumulating USD rates. This is not necessary as all rates are normalized into USD.


yotamhc (approved):
Looks good, thanks a lot!

src/xrc/src/forex.rs
```diff
@@ -356,26 +365,25 @@ impl ForexRateStore {
         Err(GetForexRateError::InvalidTimestamp(timestamp))
     }
 
-    /// Puts or updates rates for a given timestamp. If rates already exist for the given timestamp, only rates for which a new rate with higher number of sources are replaced.
-    pub(crate) fn put(&mut self, timestamp: u64, rates: ForexMultiRateMap) {
+    /// Puts or updates rates for a given timestamp. If rates already exist for the given timestamp,
+    /// only rates for which a new rate with higher number of sources are replaced.
+    pub(crate) fn put(&mut self, timestamp: u64, mut rates: ForexMultiRateMap) {
```
THLO, line 370: How about removing the USD rates before calling this function so that no mutable reference is required here?
If that's messy (for example because we need to remember to remove it wherever the function is called), we can also leave it the way you proposed.

THLO (approved):

src/xrc/src/forex.rs
dfinity-ryancroote: Good point. I moved the removal to the when the retrieval of the rates map from the collector occurs.

src/xrc/src/forex.rs
```diff
@@ -277,16 +277,18 @@ impl ForexRateStore {
...
     ) -> Result<QueriedExchangeRate, GetForexRateError> {
         // Normalize timestamp to the beginning of the day.
-        let mut timestamp = (requested_timestamp / SECONDS_PER_DAY) * SECONDS_PER_DAY;
+        let mut requested_timestamp = (requested_timestamp / SECONDS_PER_DAY) * SECONDS_PER_DAY;
 
+        // If today's date is requested, and the day is not over anywhere on Earth, use yesterday's date
+        // Get the normalized timestamp for yesterday.
         if !cfg!(feature = "disable-forex-timezone-offset") {
             // If today's date is requested, and the day is not over anywhere on Earth, use yesterday's date
             // Get the normalized timestamp for yesterday.
```
THLO, line 286: These lines should be removed, right?

src/xrc/src/forex.rs
```diff
@@ -277,16 +277,18 @@ impl ForexRateStore {
...
-            let yesterday = (current_timestamp as i64 + TIMEZONE_AOE_SHIFT_SECONDS) as u64
-                / SECONDS_PER_DAY
-                * SECONDS_PER_DAY;
-            if timestamp > SECONDS_PER_DAY && yesterday == timestamp {
-                timestamp -= SECONDS_PER_DAY;
+            // If today's date is requested, and the day is not over anywhere on Earth, use yesterday's date
+            // Get the normalized timestamp for yesterday.
```
THLO, line 288: The same sentence again...?

src/xrc/src/forex.rs
```diff
@@ -302,88 +304,95 @@ impl ForexRateStore {
...
                 let quote = rates_for_timestamp.get(&quote_asset);
 
                 match (base, quote) {
                     (Some(base_rate), Some(quote_rate)) => {
-                        return Ok(base_rate.clone() / quote_rate.clone())
+                        return Ok(base_rate.clone() / quote_rate.clone());
                     }
-                    (Some(base_rate), None) => {
+                    (Some(_), None) => {
                         // If the quote asset is USD, it should not be present in the map and the base rate already uses USD as the quote asset.
```
THLO, line 353: This line can be removed because the quote asset cannot be USD.

src/xrc/src/forex.rs
```diff
@@ -302,88 +304,95 @@ impl ForexRateStore {
...
             }
         }
         // If we got here, no rate is found for this timestamp within a range of [MAX_DAYS_TO_GO_BACK] days before it.
-        Err(GetForexRateError::InvalidTimestamp(timestamp))
+        Err(GetForexRateError::InvalidTimestamp(requested_timestamp))
     }
 
-    /// Puts or updates rates for a given timestamp. If rates already exist for the given timestamp, only rates for which a new rate with higher number of sources are replaced.
+    /// Puts or updates rates for a given timestamp. If rates already exist for the given timestamp,
+    /// only rates for which a new rate with higher number of sources are replaced.
```
THLO, line 380:
```suggestion
    /// only rates for which a new rate with a higher number of sources are replaced.
```

src/xrc/src/forex.rs
```diff
@@ -302,88 +304,95 @@ impl ForexRateStore {
...
                 }
             }
         }
         // If we got here, no rate is found for this timestamp within a range of [MAX_DAYS_TO_GO_BACK] days before it.
-        Err(GetForexRateError::InvalidTimestamp(timestamp))
+        Err(GetForexRateError::InvalidTimestamp(requested_timestamp))
     }
 
-    /// Puts or updates rates for a given timestamp. If rates already exist for the given timestamp, only rates for which a new rate with higher number of sources are replaced.
+    /// Puts or updates rates for a given timestamp. If rates already exist for the given timestamp,
```
THLO, line 379:
```suggestion
    /// Inserts or updates rates for a given timestamp. If rates already exist for the given timestamp,
```

src/xrc/src/forex.rs
```diff
@@ -1641,6 +1655,57 @@ mod test {
         assert!(!collector.update("src1".to_string(), first_day_timestamp, rates));
     }
 
+    #[test]
+    fn ensure_rate_store_removes_usd_rates() {
+        let mut collector = ForexRatesCollector::new();
+
+        // Start by executing the same logic as for the [OneDayRatesCollector] to verify that the calls are relayed correctly
```
THLO, line 1662:
```suggestion
        // Start by executing the same logic as for the [OneDayRatesCollector] to verify that the calls are relayed correctly.
```

src/xrc/src/forex.rs
```diff
@@ -1795,6 +1883,140 @@ mod test {
         );
     }
 
+    #[test]
+    fn rate_store_gets_rate_in_past_if_current_day_is_not_over() {
+        // Create a store, update, check that only rates with more sources were updated.
+        let mut store = ForexRateStore::new();
+        // Rate 0
```
THLO, line 1890: Shouldn't it be "Day 0"?

src/xrc/src/forex.rs
```diff
@@ -1795,6 +1883,140 @@ mod test {
...
+                            symbol: USD.to_string(),
+                            class: AssetClass::FiatCurrency,
+                        },
+                        timestamp: 0,
+                        rates: vec![800_000_000],
+                        base_asset_num_queried_sources: 4,
+                        base_asset_num_received_rates: 4,
+                        quote_asset_num_queried_sources: 4,
+                        quote_asset_num_received_rates: 4,
+                        forex_timestamp: Some(1234),
```
THLO, line 1910: This is not a problem here but we don't want forex timestamps in the future.

src/xrc/src/forex.rs
```diff
@@ -1795,6 +1883,140 @@ mod test {
...
+
+        // If the current timestamp is at day 2 and the requested timestamp is at day 2,
+        // return the rate for day 1 as day 2 is still active anywhere on Earth.
+        let result = store.get(SECONDS_PER_DAY * 2, SECONDS_PER_DAY * 2, "EUR", USD);
+        assert!(matches!(
+            result,
+            Ok(rate) if rate.rates == vec![1_000_000_000] && rate.base_asset_num_received_rates == 5,
+        ));
+
+        // If the current timestamp is at start of day 2 for UTC -12 and the
```
THLO, line 1987:
```suggestion
        // If the current timestamp is at start of day 2 for UTC-12 and the
```

THLO (approved):

src/xrc/src/forex.rs
```diff
@@ -277,16 +277,18 @@ impl ForexRateStore {
...
             // If today's date is requested, and the day is not over anywhere on Earth, use yesterday's date
             // Get the normalized timestamp for yesterday.
-            let yesterday = (current_timestamp as i64 + TIMEZONE_AOE_SHIFT_SECONDS) as u64
-                / SECONDS_PER_DAY
-                * SECONDS_PER_DAY;
-            if timestamp > SECONDS_PER_DAY && yesterday == timestamp {
-                timestamp -= SECONDS_PER_DAY;
+            // If today's date is requested, and the day is not over anywhere on Earth, use yesterday's date
+            // Get the normalized timestamp for yesterday.
+            let aoe_end = requested_timestamp.saturating_add(TIMEZONE_AOE_SHIFT_SECONDS);
```
yotamhc, line 289: I keep getting confused here, as when seeing `aoe_end` I think it means "today's end AOE", but it actually means "requested timestamp's day end AOE". So maybe rename it to `requested_day_aoe_end`?
(I know it's long and ugly, but may prevent future confusion)

yotamhc (approved):

src/xrc/src/forex.rs
dfinity-ryancroote: Removed.

src/xrc/src/forex.rs
dfinity-ryancroote: Fixed the timestamps to be the same as the day.

PR #185, opened 2023-03-13
Changing the CXDR computation

THLO:
This PR changes the way the artificial CXDR rates are computed:
Rather than using `[median_rate-d, median_rate, median_rate+d]`, choosing `d` such that we get the right variance, the rates are now `[median_rate, median_rate, median_rate+d']` for a `d'` that results in the same variance.

As a result, we can no longer get a rate `0` when the variance is large.

### DIFF
--- a/src/xrc/src/forex.rs
+++ b/src/xrc/src/forex.rs
@@ -502,7 +502,11 @@ impl OneDayRatesCollector {
                 .saturating_add(weighted_gbp_std_dev.saturating_pow(2)))
             .saturating_div(1_000_000_000_000); // Removing the factor (10^6)^2 due to the weight scaling.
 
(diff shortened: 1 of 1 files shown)


src/xrc/src/forex.rs
```diff
@@ -1914,6 +1914,44 @@ mod test {
         assert_eq!(cxdr_usd_rate, _expected_rate);
     }
 
+    /// This function tests that the computed set of articial CXDR rates does not contain any zero rates.
```
dfinity-ryancroote, line 1917:
```suggestion
    /// This function tests that the computed set of artificial CXDR rates does not contain any zero rates.
```

dfinity-ryancroote (approved):

PR #186, opened 2023-03-14
`bugfix`: Check the division when inverting the rate to avoid division by zero errors.

dfinity-ryancroote:
This PR resolves the discovered possibility of division by zero errors while inverting a rate. This is accomplished by updating the division in `utils::invert_rate` (renamed to `utils::checked_invert_rate`) to `checked_div` and returning the result.

* When used to invert a `QueriedExchangeRate`, the failed results are filtered out using `filter_map`.
* When inverting a stablecoin rate, if the rate fails to be inverted, a `CallExchangeError::NoRateFound` is returned.

### DIFF
--- a/src/xrc/src/lib.rs
+++ b/src/xrc/src/lib.rs
@@ -279,7 +279,9 @@ impl std::ops::Mul for QueriedExchangeRate {
                     .saturating_mul(other_value)
                     .saturating_div(RATE_UNIT as u128) as u64;
 
-                rates.push(rate);
+                if rate > 0 {
+                    rates.push(rate);
+                }
             }
         }
         rates.sort();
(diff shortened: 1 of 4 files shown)


src/xrc/src/lib.rs
```diff
@@ -380,7 +380,7 @@ impl QueriedExchangeRate {
         let mut inverted_rates: Vec<_> = self
             .rates
             .iter()
-            .map(|rate| utils::invert_rate(*rate))
+            .filter_map(|rate| utils::checked_invert_rate(*rate))
```
THLO, line 383: How about adding a small test where a rate is inverted (or divided), showing that the zero rates are filtered out as expected?

THLO (approved):

src/xrc/src/lib.rs
dfinity-ryancroote: Meant to add one. 😓 

Test has been added. Found that we should filter out zero rates in the multiplication as well.

THLO: Why is it needed there as well? There's no division in multiplication (other than dividing by `RATE_UNIT`).
Sure, multiplication will increase the number of zeroes but that's not problematic. In fact, when we remove the zeroes, they are potentially excluded from the variance.
Or am I missing something?

dfinity-ryancroote: Ah, I thought the plan was to remove the zeros completely. I'll revert that.

dfinity-ryancroote: Reverted and updated the test.

PR #187, opened 2023-03-14
`feat`: Requested timestamps that are too far in the future should be rejected and charged the minimum fee.

dfinity-ryancroote:
This PR introduces a future timestamp check for rate requests. If the timestamp is too far in the future, it is rejected and the minimum fee will be charged.

### DIFF
--- a/src/xrc/src/api.rs
+++ b/src/xrc/src/api.rs
@@ -5,7 +5,7 @@ mod test;
 pub use metrics::get_metrics;
 
 use crate::cache::ExchangeRateCache;
-use crate::candid::OtherError;
+use crate::errors;
 use crate::{
     call_exchange,
     candid::{Asset, AssetClass, ExchangeRateError, GetExchangeRateRequest, GetExchangeRateResult},
@@ -376,7 +376,7 @@ async fn handle_cryptocurrency_pair(
 
         env.charge_cycles(charge_cycles_option)?;
         if timestamp_is_in_future {
-            return Err(timestamp_is_in_future_error(
+            return Err(errors::timestamp_is_in_future_error(
                 timestamp.value,
                 current_timestamp,
             ));
@@ -507,7 +507,7 @@ async fn handle_crypto_base_fiat_quote_pair(
 
         env.charge_cycles(charge_cycles_option)?;
         if timestamp_is_in_future {
-            return Err(timestamp_is_in_future_error(
+            return Err(errors::timestamp_is_in_future_error(
                 timestamp.value,
                 current_timestamp,
             ));
@@ -604,7 +604,7 @@ fn handle_fiat_pair(
         .map_err(|err| err.into())
         .and_then(validate)
     } else {
-        Err(timestamp_is_in_future_error(
+        Err(errors::timestamp_is_in_future_error(
             requested_timestamp,
             current_timestamp,
         ))
@@ -725,17 +725,4 @@ async fn call_exchange_for_stablecoin(
     } else {
         result
     }
-}
-
-fn timestamp_is_in_future_error(
-    requested_timestamp: u64,
-    current_timestamp: u64,
-) -> ExchangeRateError {
-    ExchangeRateError::Other(OtherError {
-        code: 1,
-        description: format!(
-            "Current IC time is {}. {} is in the future!",
-            current_timestamp, requested_timestamp
-        ),
-    })
 }
--- a/src/xrc/src/api.rs
+++ b/src/xrc/src/api.rs
@@ -5,6 +5,7 @@ mod test;
 pub use metrics::get_metrics;
 
 use crate::cache::ExchangeRateCache;
+use crate::candid::OtherError;
 use crate::{
     call_exchange,
     candid::{Asset, AssetClass, ExchangeRateError, GetExchangeRateRequest, GetExchangeRateResult},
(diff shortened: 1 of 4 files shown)


src/xrc/src/api.rs
```diff
@@ -356,19 +357,31 @@ async fn handle_cryptocurrency_pair(
     }
 
     if !utils::is_caller_privileged(&caller) {
+        let current_timestamp = env.time_secs();
+        let timestamp_is_in_future = timestamp.value > current_timestamp;
```
THLO, line 361: I think we can be a bit more lenient: If the timestamp maps to the current minute, it is still okay:

`timestamp_is_in_future = timestamp.value / 60 > current_timestamp / 60`

THLO, line 361: Does it make sense to allow a privileged canister to request a timestamp in the future?
If it results in simpler code, then the answer is probably "yes"...

src/xrc/src/api.rs
```diff
@@ -476,18 +489,30 @@ async fn handle_crypto_base_fiat_quote_pair(
...
             ChargeOption::OutboundRatesNeeded(num_rates_needed)
         };
 
         env.charge_cycles(charge_cycles_option)?;
+        if timestamp_is_in_future {
+            return Err(errors::timestamp_is_in_future_error(
+                timestamp.value,
+                current_timestamp,
+            ));
+        }
```
THLO, line 514: There is quite a bit of duplicate code here...

dfinity-ryancroote: I'm going to resolve this in a separate PR.

src/xrc/src/api.rs
```diff
@@ -565,18 +590,25 @@ fn handle_fiat_pair(
     env: &impl Environment,
     request: &GetExchangeRateRequest,
 ) -> Result<QueriedExchangeRate, ExchangeRateError> {
-    let timestamp = utils::get_normalized_timestamp(env, request);
+    let requested_timestamp = utils::get_normalized_timestamp(env, request);
     let current_timestamp = env.time_secs();
-    let result = with_forex_rate_store(|store| {
-        store.get(
-            timestamp,
+    let result = if requested_timestamp <= current_timestamp {
```
THLO, line 595: Same here.

src/xrc/src/api/test.rs
```diff
@@ -1527,3 +1527,106 @@ fn cached_rate_with_few_collected_rates_is_ignored_for_privileged_canister() {
...
+    use crate::{candid::OtherError, errors::TIMESTAMP_IS_IN_FUTURE_ERROR_CODE, ONE_MINUTE};
+
+    use super::*;
+
+    /// This function tests that a crypto pair request with a timestamp in the future
+    /// is rejected and charged the minimum fee.
+    #[test]
+    fn handle_cryptocurrency_pair() {
+        let current_timestamp: u64 = 1678752000;
+        let future_timestamp = current_timestamp.saturating_add(2 * ONE_MINUTE);
```
THLO, line 1542: Adding one minute would be enough, right?

src/xrc/src/api/test.rs
```diff
@@ -1527,3 +1527,106 @@ fn cached_rate_with_few_collected_rates_is_ignored_for_privileged_canister() {
...
+            matches!(result, Err(ExchangeRateError::Other(OtherError { code, description: _ })) if code == TIMESTAMP_IS_IN_FUTURE_ERROR_CODE)
+        );
+    }
+
+    /// This function tests that a crypto/fiat pair request with a timestamp in the future
+    /// is rejected and charged the minimum fee.
+    #[test]
+    fn handle_crypto_base_fiat_quote_pair() {
+        let current_timestamp: u64 = 1678752000;
+        let future_timestamp = current_timestamp.saturating_add(2 * ONE_MINUTE);
```
THLO, line 1574: Same here.

src/xrc/src/api/test.rs
```diff
@@ -1527,3 +1527,106 @@ fn cached_rate_with_few_collected_rates_is_ignored_for_privileged_canister() {
...
+            matches!(result, Err(ExchangeRateError::Other(OtherError { code, description: _ })) if code == TIMESTAMP_IS_IN_FUTURE_ERROR_CODE)
+        );
+    }
+
+    /// This function tests that a fiat pair request with a timestamp in the future
+    /// is rejected and charged the minimum fee.
+    #[test]
+    fn handle_fiat_pair() {
+        let current_timestamp: u64 = 1678752000;
+        let future_timestamp = current_timestamp.saturating_add(2 * ONE_MINUTE);
```
THLO, line 1606: And here.   :-)

THLO (approved):

src/xrc/src/api.rs
dfinity-ryancroote: The code is simpler, but it would not be much too much work. When I work on the duplicate code, I'll address this as well.

dfinity-ryancroote: I don't think this is needed as `timestamp.value` is a normalized timestamp to the latest minute, so if it is still greater than the `current_timestamp` then `timestamp` is in the future.

src/xrc/src/api.rs
dfinity-ryancroote: `requested_timestamp` is normalized to the latest minute. It should not be larger than the `current_timestamp` at this point.

src/xrc/src/api.rs
THLO: You are right. If the timestamp is normalized, the seconds are removed. If it is still larger than current_timestamp, it must be the next minute or later. So, dividing by 60 is not necessary.

PR #190, opened 2023-03-16
`refactor`: Clean up API code that was duplicated.

dfinity-ryancroote:
This PR cleans up code in the API that was duplicated in the various `handle_*` functions.

* Creates a `validate_request` function to check for future timestamps, rate limiting, already inflight calls, and past timestamps not found in the rate cache.
* Creates a `charge_cycles` function to handle the basic logic the occurs in the `handle_*` functions for charging cycles.
* Adds a the following test: privileged callers cannot make requests containing timestamps in the future.

### DIFF
--- a/src/xrc/src/api.rs
+++ b/src/xrc/src/api.rs
@@ -334,6 +334,73 @@ fn get_rate_from_cache(
     }
 }
 
+/// The possible errors that [validate_request] may return if a request
+/// fails the validation.
+enum ValidateRequestError {
+    /// The timestamp is in the future.
+    FutureTimestamp {
+        /// The normalized requested timestamp.
+        requested_timestamp: u64,
+        /// The current IC time in seconds.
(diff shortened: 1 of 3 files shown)


src/xrc/src/api.rs
```diff
@@ -334,18 +335,117 @@ fn get_rate_from_cache(
...
+}
+
+/// This function attempts to charge cycles for the request being made.
+/// If an error is found validating the request, charge the minimum fee.
+/// Otherwise, charge what is necessary for the number of needed outbound calls.
+/// If the caller is privileged, exit early as they are not charged cycles.
+fn charge_cycles(
+    env: &impl Environment,
+    num_rates_needed: usize,
+    error_found_validating_request: bool,
```
THLO, line 412: How about something shorter like `is_valid_request`?

THLO (approved):
Very nice!

PR #194, opened 2023-03-29
bugfix: Add a updating forex store guard for safety purposes

dfinity-ryancroote:
This PR adds a guard to the periodic task for updating the forex store. While we have been extremely careful to dealing with panics, this guard better ensures the state is set properly in case we have missed something.


THLO:
While I like the introduction of a guard, can we have a test that shows the value of this addition?

THLO (approved):

PR #199, opened 2023-04-05
feat: Add dashboard to view most recent requests

dfinity-ryancroote:
This PR adds a dashboard to the canister. This will give us a better view into how the XRC is being used.

* A metadata table to see the following:
  * Decimals
  * Number of exchanges available
  * Number of forex sources available
  * List of the privileged canister IDs
* Tables with information on the active forex collectors:
  * Sources that have been collected
* Tables containing requests and responses including the timestamp and caller

### DIFF
--- a/src/xrc/src/forex.rs
+++ b/src/xrc/src/forex.rs
@@ -571,6 +571,10 @@ impl ForexRatesCollector {
         }
     }
 
+    pub(crate) fn get_timestamps(&self) -> Vec<u64>  {
+        self.days.iter().map(|day| day.timestamp).collect()
+    }
+
     /// Updates the collected rates with a new set of rates. The provided timestamp must exist in the collector or be newer than the existing ones. The function returns true if the collector has been updated, or false if the timestamp is too old.
     pub(crate) fn update(&mut self, source: String, timestamp: u64, rates: ForexRateMap) -> bool {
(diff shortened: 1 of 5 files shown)


src/xrc/src/forex.rs
```diff
@@ -571,6 +571,10 @@ impl ForexRatesCollector {
         }
     }
 
+    pub(crate) fn get_timestamps(&self) -> Vec<u64>  {
```
THLO, line 574: Missing documentation.   :-)

src/xrc/src/request_log.rs
```diff
@@ -0,0 +1,59 @@
+use std::{cell::RefCell, collections::VecDeque, thread::LocalKey};
+
+use candid::Principal;
+use ic_xrc_types::{GetExchangeRateRequest, GetExchangeRateResult};
+
+pub(crate) struct RequestLogEntry {
```
THLO, line 6: Could you please add some documentation to this file as well?

THLO (approved):

src/xrc/src/forex.rs
dfinity-ryancroote: Added!

src/xrc/src/request_log.rs
dfinity-ryancroote: Added!

PR #202, opened 2023-04-10
`feat`: Adds additional validation for incoming request's asset symbols.

dfinity-ryancroote:
This PR adds validation for the symbols. If the symbols contain invalid symbols, the sanitize function clears the field. The validation picks up on the empty strings and causes an error to be returned. This is added to prevent calls to the exchanges for invalid symbols.


src/xrc/src/api.rs
```diff
@@ -396,6 +399,10 @@ enum ValidateRequestError {
     AlreadyInflight,
     /// The request timestamp that goes back in the past is not in the rate cache.
     PastTimestampNotCached,
+    CryptoBaseAssetEmptySymbol,
+    FiatBaseAssetEmptySymbol,
+    CryptoQuoteAssetEmptySymbol,
+    FiatQuoteAssetEmptySymbol,
```
THLO, line 405: Is it useful to distinguish between `CryptoBaseAssetEmptySymbol` and `FiatBaseAssetEmptySymbol`?
Wouldn't `BaseAssetEmptySymbol` and `QuoteAssetEmptySymbol` be enough?

src/xrc/src/api/test.rs
```diff
@@ -1662,3 +1662,295 @@ mod timestamp_is_in_future {
...
+            .with_cycles_available(XRC_REQUEST_CYCLES_COST)
+            .with_accepted_cycles(XRC_MINIMUM_FEE_COST)
+            .build();
+        let request = GetExchangeRateRequest {
+            base_asset: Asset {
+                symbol: "ICP".to_string(),
+                class: AssetClass::Cryptocurrency,
+            },
+            quote_asset: Asset {
+                symbol: "<>".to_string(),
```
THLO, line 1719: How about using different invalid symbols in each test?

Let's be creative (/ç%^*@ßðæđßħłĸ¶ł«»¢nµþœŧ€đŋ...)!   :-)

src/xrc/src/api/test.rs
```diff
@@ -1662,3 +1662,295 @@ mod timestamp_is_in_future {
...
+        let result = get_exchange_rate_internal(&env, &call_exchanges_impl, &request)
+            .now_or_never()
+            .expect("future should complete");
+        assert!(matches!(
+            result,
+            Err(ExchangeRateError::CryptoQuoteAssetNotFound)
+        ));
+    }
+
+    /// This function tests that a crypto/fiat pair request with an invalid base asset symbol
```
THLO, line 1734: The asset class doesn't have any effect here. Is there a benefit in testing all possible scenarios?

THLO (approved):

src/xrc/src/api.rs
dfinity-ryancroote: It's useful in the sense of mapping the error to the `CryptoBaseAssetNotFound`, etc. If we want, we could simply this down to just use the `Other` error type and adding a new error type. It just seemed friendlier to match up the errors.

src/xrc/src/api/test.rs
dfinity-ryancroote: To ensure total coverage, we have to go through all scenarios. The `validate_request` function is used in each `handle_` function. At some point, I would like to figure out a way to refactor it to be in a singular point, but I would also like to avoid having many places where we could charge for cycles.

THLO (approved):

PR #203, opened 2023-04-12
`feat`: Add new forex source: Reserve Bank of Australia

dfinity-ryancroote:
This PR adds a new forex source: Reserve Bank of Australia.

### DIFF
--- a/src/xrc/src/periodic.rs
+++ b/src/xrc/src/periodic.rs
@@ -477,7 +477,7 @@ mod test {
         let timestamp = 1680220800;
         let forexes_with_timestamps_and_context =
             get_forexes_with_timestamps_and_context(timestamp);
-        assert_eq!(forexes_with_timestamps_and_context.len(), 6);
+        assert_eq!(forexes_with_timestamps_and_context.len(), 7);
 
         assert!(matches!(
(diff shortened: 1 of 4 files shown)


yotamhc (approved):
I like the idea of separating the sources into different modules! We should do that with the existing ones as well.

src/xrc/src/periodic.rs
```diff
@@ -536,5 +536,15 @@ mod test {
...
             1680134400
         );
+
+        assert!(matches!(
+            forexes_with_timestamps_and_context[6].0,
+            Forex::ReserveBankOfAustralia(_)
+        ));
+        assert_eq!(forexes_with_timestamps_and_context[5].1, 1680134400);
+        assert_eq!(
+            forexes_with_timestamps_and_context[5].2.timestamp,
```
nategurian, line 546: Should these be:
`forexes_with_timestamps_and_context[6].1` and `forexes_with_timestamps_and_context[6].2`

dfinity-ryancroote: Good catch. Fixed.

nategurian (approved):

gauravrai-py (approved):

PR #207, opened 2023-04-14
`refactor`: Consolidate time-related constants.

dfinity-ryancroote:
This PR consolidates some similar constants that we had in the project related to time.


src/xrc/src/lib.rs
```diff
@@ -101,6 +101,10 @@ const ONE_KIB: u64 = 1_024;
 
 /// 1 minute in seconds
 const ONE_MINUTE: u64 = 60;
+// 1 hour in seconds
+const ONE_HOUR: u64 = 60 * ONE_MINUTE;
+// 1 day in seconds
+const ONE_DAY: u64 = 24 * ONE_HOUR;
```
nategurian, line 107: This is just a suggestion, I am not sure if it really would make sense. You replace SECONDS_PER_HOUR and SECONDS_PER_DAY with ONE_HOUR and ONE_DAY. You have in the comments that it seconds, but does it make sense to possibly change the names to ONE_HOUR_SECONDS or ONE_DAY_SECONDS (or something to suggest the unit)

dfinity-ryancroote: Good point. Updated.

nategurian (approved):

PR #210, opened 2023-04-14
Validating the collected rates

THLO:
This PR changes the way the `QueriedExchangeRate` struct handles the collected rates.
In particular, it verifies that the rates lie within a range of values that are deemed valid. If half or more of the received rates are invalid, all rates are dropped.
The median computation is also adapted as a result of this change, ensuring that less than half of the rates can be off arbitrarily - even for cryptocurrency pairs - and the returned rate is still valid.

Since the `QueriedExchangeRate` struct is no more strict in assessing the validity of the given data in the constructor, several tests needed to be adapted.

Finally, the PR introduces several convenience functions to define assets (which could have been done in a separate PR but it was convenient to do it in one go).


src/xrc/src/lib.rs
```diff
@@ -332,6 +336,36 @@ impl std::ops::Mul for QueriedExchangeRate {
...
+                    .saturating_mul(other_rate as u128)
+                    .saturating_div(RATE_UNIT as u128);
+                if product_rate <= (RATE_UNIT * RATE_UNIT) as u128 {
+                    Some(product_rate as u64)
+                } else {
+                    None
+                }
+            }
+            (_, _) => None,
+        };
```
dfinity-ryancroote, line 351: There is a significant problem. The forex rates that we have been storing into the store will not have the `median_rate` field set after the upgrade basically invalidating them. I have two ideas:

1. We can update the values in the post_upgrade.
2. We can update the median rate on the fly when accessed.

src/xrc/src/lib.rs
```diff
@@ -402,22 +440,40 @@ impl AllocatedBytes for Vec<u64> {
...
-                standard_deviation: standard_deviation(&rate.rates),
-                forex_timestamp: rate.forex_timestamp,
-            },
+impl TryFrom<QueriedExchangeRate> for ExchangeRate {
+    type Error = ExchangeRateError;
+
+    fn try_from(rate: QueriedExchangeRate) -> Result<Self, Self::Error> {
+        match rate.median_rate {
+            Some(median_rate) => {
+                // If no rates were invalid,the median over all rates can be used.
```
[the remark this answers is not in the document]
THLO: The logic is the following: We only return a rate if the `median_rate` field is set. In other words, if it is not set, something went wrong, resulting in an error.
The `median_rate`, however, might actually be a function of multiple medians, for example the product of two medians.
This is not the same thing as the median over the product of rates. That's where the conditional check comes in: If all rates were valid, it is safe to compute the median over the product of rates for a more accurate result.

src/xrc/src/lib.rs
THLO: This is indeed a significant problem. We definitely have to solve this problem before merging this PR.

src/xrc/src/lib.rs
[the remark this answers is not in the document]
yotamhc: But what guarantees that `median_rate` that is picked in line 455 is valid? (in the case the condition does not hold)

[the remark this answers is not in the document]
THLO: There is code in the constructor to ensure that `median_rate` is valid (i.e., more than half of the received values are valid). If rates are dropped, we use the product of two medians (which are both valid individually) instead of the median of the cross product of rates.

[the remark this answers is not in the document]
yotamhc: Sorry, I still don't get something here (and other than convincing me, it might be good to make the comment clear enough to convince other future readers)
If `median_rate` is ensured to be valid, why don't we just use it in all cases?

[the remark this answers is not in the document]
THLO: I can explain the rationale face-to-face but the design is likely to change quite a bit because of the compatibility problem that @dfinity-ryancroote raised.

If we remove `median_rate` again, there is no need to add some elaborate comments now.   :-)

src/xrc/src/lib.rs
THLO: The latest changes should fix this problem. Please take another look!

src/xrc/src/lib.rs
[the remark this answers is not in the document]
THLO: The median_rate is now gone. The new mechanism should be easier to understand. @yotamhc, please take another look!

src/xrc/src/lib.rs
```diff
@@ -486,6 +504,23 @@ impl QueriedExchangeRate {
...
+            description: INVALID_RATE_ERROR_MESSAGE.to_string(),
+        }));
+    }
+    if rate.is_valid() {
+        Ok(rate)
+    } else {
+        Err(ExchangeRateError::InconsistentRatesReceived)
+    }
+}
+
```
dfinity-ryancroote, line 523: Could we combine `is_valid` and this `validate` function? Perhaps something like....

```rust
impl QueriedExchangeRate {
    fn validate(self) -> Result<Self, ExchangeRateError> {
       // validation logic here
   }
}
```

THLO: Yeah, I had this idea as well.
Let me try it out!

THLO: Done! Please take another look!

src/xrc/src/lib.rs
```diff
@@ -470,19 +488,28 @@ impl QueriedExchangeRate {
...
         for end in diff..num {
             if self.rates[end] - self.rates[end - diff]
                 <= self.rates[end - diff].saturating_div(RATE_DEVIATION_DIVISOR)
             {
-                return true;
+                return Ok(self);
             }
         }
-        false
+        Err(ExchangeRateError::InconsistentRatesReceived)
```
dfinity-ryancroote, line 512: I'm not a big fan of have the return on line 509. I would expect the `Ok` result would be the last line. I would suggest the following:

```rust
        if (diff..num).all(|end| {
            self.rates[end] - self.rates[end - diff]
                > self.rates[end - diff].saturating_div(RATE_DEVIATION_DIVISOR)
        }) {
            return Err(ExchangeRateError::InconsistentRatesReceived);
        }

        Ok(self)
```

THLO: I agree, it looks nicer. Thanks for providing the code snippet! That made it very easy to change the code.   :-)

PR #217, opened 2023-06-01
feat: ready to be pulled

lwshang:
This PR will make exchange rate canister the first "pullable" mainnet canister.

<details>
  <summary>Click Me: What does "pullable" mean?</summary>
  
 Once the exchange rate canister is deployed, any app canister integrating with the exchange rate canister will only need specify its dependency in `dfx.json` as:

```json
{
  "canisters": {
    "app": {
      "type": "motoko",
      "main": "src/main.mo",
      "dependencies": [
        "exchange-rate"
      ]
    },
    "exchange-rate": {
      "type": "pull",
      "id": "uf6dk-hyaaa-aaaaq-qaaaq-cai"
    }
}
```

Then they can pull the dependency from the mainnet and deploy it locally:

```sh
dfx deps pull    # fetch the metadata and download the wasm
dfx deps init    # setup init arguments (exchange rate canister needs no argument)
dfx deps deploy  # deploy on the local replica with the same canister ID as mainnet
```

Check this [page](https://github.com/dfinity/sdk/blob/master/docs/concepts/pull-dependencies.md) for an instruction of "pull dependencies".
</details>

# Changes

The major change in this PR is in the `dfx.json`. It adds the necessary metadata in the "pullable" field.

Also `dfx` handles the wasm optimization and gzip during canister building.

# Next Step

One important thing to make the canister "pullable" is serving the wasm module through an URL as the `wasm_url` field in `dfx.json`.

I entered "https://github.com/dfinity/exchange-rate-canister/releases/latest/download/xrc.wasm.gz" which doesn't exist yet.

In this way, we will use GitHub Releases to host the wasm. The workflow would be:
* Deploy the canister on mainnet
* Make a new GitHub Release which attached the **exact** wasm module

A common practice is creating a GItHub Action workflow triggered by git tag.


dfx.json
```diff
@@ -1,12 +1,19 @@
...
       "candid": "./src/xrc/xrc.did",
       "build": "./scripts/build-wasm",
-      "wasm": "./target/wasm32-unknown-unknown/release/xrc.wasm"
+      "wasm": "./target/wasm32-unknown-unknown/release/xrc.wasm",
+      "optimize": "cycles",
+      "gzip": true,
+      "pullable": {
+        "dependencies": [],
+        "init_guide": "",
+        "wasm_url": "https://github.com/dfinit/exchange-rate-canister/releases/latest/download/xrc.wasm.gz"
```
dfinity-ryancroote, line 15:
```suggestion
        "wasm_url": "https://github.com/dfinity/exchange-rate-canister/releases/latest/download/xrc.wasm.gz"
```

.github/workflows/tests.yml
```diff
@@ -6,6 +6,9 @@ on:
       - main
   pull_request:
 
+env:
+  DFX_VERSION: 0.14.1-beta.1
```
dfinity-ryancroote, line 10:
```suggestion
  DFX_VERSION: 0.14.1
```

dfinity-ryancroote (approved):

PR #225, opened 2023-06-15
test: Add caching `e2e` test

dfinity-ryancroote:
This PR adds an e2e test for the crypto caching mechanism.

---

Setup:
* Deploy mock FOREX data providers and exchanges with a large response delay

Start replicas and deploy the XRC, configured to use the mock data sources

Runbook:

1. Request the same exchange rate many times per second for a bounded time, e.g., 1 minute
2. Assert that all requests are answered with the same rate with a small delay after the first response (due to caching in the XRC)

Success criteria:

* All queries are handled correctly

### DIFF
--- a/src/xrc-tests/src/tests/caching.rs
+++ b/src/xrc-tests/src/tests/caching.rs
@@ -24,7 +24,7 @@ use crate::{
 ///
 /// * All queries are handled correctly
 ///
-/// How are the expected values determined:
+/// The expected values are determined as follows:
 ///
 /// Crypto-pair (retrieve ICP/BTC rate)
 /// 0. The XRC retrieves the ICP/USDT rate.
--- a/src/xrc-tests/src/tests/caching.rs
+++ b/src/xrc-tests/src/tests/caching.rs
@@ -1,6 +1,9 @@
 use std::time::Instant;
 
-use ic_xrc_types::{Asset, AssetClass, GetExchangeRateRequest, GetExchangeRateResult};
+use ic_xrc_types::{
+    Asset, AssetClass, ExchangeRate, ExchangeRateMetadata, GetExchangeRateRequest,
+    GetExchangeRateResult,
+};
 
 use crate::{
     container::{run_scenario, Container},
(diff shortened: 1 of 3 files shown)


src/xrc-tests/src/tests/caching.rs
```diff
@@ -0,0 +1,141 @@
...
+/// Runbook:
+///
+/// 1. Request the same exchange rate many times per second for a bounded time, e.g., 1 minute
+/// 2. Assert that all requests are answered with the same rate with a small delay after the first response (due to caching in the XRC)
+///
+/// Success criteria:
+///
+/// * All queries are handled correctly
+///
+/// How are the expected values determined:
```
THLO, line 24:
```suggestion
/// The expected values are determined as follows:
```

src/xrc-tests/src/tests/caching.rs
```diff
@@ -0,0 +1,141 @@
...
+                Ok(exchange_rate) => {
+                    assert_eq!(exchange_rate.base_asset, request.base_asset);
+                    assert_eq!(exchange_rate.quote_asset, request.quote_asset);
+                    assert_eq!(exchange_rate.timestamp, timestamp_seconds);
+                    assert_eq!(exchange_rate.metadata.base_asset_num_queried_sources, 7);
+                    assert_eq!(exchange_rate.metadata.base_asset_num_received_rates, 7);
+                    assert_eq!(exchange_rate.metadata.quote_asset_num_queried_sources, 7);
+                    assert_eq!(exchange_rate.metadata.quote_asset_num_received_rates, 7);
+                    assert_eq!(exchange_rate.metadata.standard_deviation, 3_483_761);
+                    assert_eq!(exchange_rate.rate, 88_587_570);
```
THLO, line 120: Alternatively, you could create an `expected_exchange_rate` and just compare for equality of the whole struct (but I'm not sure it would be any better).

THLO (approved):

src/xrc-tests/src/tests/caching.rs
```diff
@@ -0,0 +1,141 @@
...
+                }
+                Err(error) => panic!("Received an error from the XRC: {:?}", error),
+            };
+        }
+
+        // Compare the response times of the samples to ensure the cache mechanism was used
+        // for subsequent requests after the first.
+        //
+        // The subsequent requests should arrive in at least half the time.
```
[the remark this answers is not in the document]
dfinity-ryancroote: The problem is dependent on the worker setup. On my devenv, it's significantly less, but on a GitHub worker, it can be significantly less than half to around half.

src/xrc-tests/src/tests/caching.rs
THLO: It derives `PartialEq`, right? Isn't that enough?

dfinity-ryancroote: It does not. You may be thinking of `QueriedExchangeRate`.

src/xrc-tests/src/tests/caching.rs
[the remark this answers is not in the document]
THLO: Okay, that's unfortunate. Since you said that it can be around half, is there a risk that the test is flaky?

src/xrc-tests/src/tests/caching.rs
dfinity-ryancroote: You're not missing anything. I completely missed the `PartialEq`. 🤦

PR #226, opened 2023-06-16
test: Add misbehavior `e2e` test

dfinity-ryancroote:
This PR adds the misbehavior e2e test.

---

Setup:
* Deploy mock FOREX data providers and exchanges, some of which are configured to be malicious
* Start replicas and deploy the XRC, configured to use the mock data sources

Runbook:
* Request exchange rate for various cryptocurrency and fiat currency pairs
* Assert that the returned rates correspond to the expected values and that the confidence is lower due to the erroneous responses

Success criteria:
* All queries return the expected values

---

This PR also adds the flexibility to set the rates for a forex mock response.


src/xrc-tests/src/mock_responses/forex/nepal.rs
```diff
@@ -41,44 +43,44 @@ pub fn build_response_body(timestamp: u64) -> ResponseBody {
                                 "name": "U.S. Dollar",
                                 "unit": 1
                             },
-                            "buy": "125.05",
-                            "sell": "125.65"
+                            "buy": rates.get("USD_BUY").cloned().unwrap_or("125.05"),
+                            "sell": rates.get("USD_SELL").cloned().unwrap_or("125.65")
```
THLO, line 47: Doesn't the `rates` map only have the key "USD"?

src/xrc-tests/src/mock_responses/forex/nepal.rs
```diff
@@ -41,44 +43,44 @@ pub fn build_response_body(timestamp: u64) -> ResponseBody {
...
                         {
                             "currency": {
                                 "iso3": "EUR",
                                 "name": "European Euro",
                                 "unit": 1
                             },
-                            "buy": "132.37",
-                            "sell": "133.00"
+                            "buy": rates.get("EUR_BUY").cloned().unwrap_or("132.37"),
+                            "sell": rates.get("EUR_BUY").cloned().unwrap_or("133.00")
```
THLO, line 56: Shouldn't it be just "EUR" or - if these keys exist - "EUR_BUY" and "EUR_SELL"?
The same question applies to the currencies below.

src/xrc-tests/src/tests/misbehavior.rs
```diff
@@ -0,0 +1,387 @@
...
+///
+/// Success criteria:
+/// * All queries return the expected values
+///
+/// The expected values are determined as follows:
+///
+/// Crypto-pair (retrieve ICP/BTC rate)
+/// 0. The XRC retrieves the ICP/USDT rate.
+///     a. ICP/USDT rates: [3900000000, 3900000000, 3911000000, 4005000000]
+///         i. There are only 4 rates as the other 3 have been filtered out as they were greater
```
THLO, line 44: This sounds like rates greater than the median rate are always filtered, making the median rate the largest rate among the remaining rates.

This is not quite accurate, right?

src/xrc-tests/src/tests/misbehavior.rs
```diff
@@ -0,0 +1,387 @@
...
+        };
+
+        let fiat_pair_result = container
+            .call_canister::<_, GetExchangeRateResult>("get_exchange_rate", &fiat_pair_request)
+            .expect("Failed to call canister for rates");
+        let exchange_rate =
+            fiat_pair_result.expect("Failed to retrieve an exchange rate from the canister.");
+        assert_eq!(exchange_rate, expected_fiat_pair_rate);
+        // Compare the rate's standard deviation to the common (non-malicious) data set.
+        // The non-malicious dataset should be shown that it has lower variability than the malicious set
```
THLO, line 380:
```suggestion
        // The non-malicious dataset should have a lower variability than the malicious set
```

src/xrc-tests/src/tests/misbehavior.rs
```diff
@@ -0,0 +1,387 @@
...
+        // provided above.
+        assert!(FIAT_PAIR_COMMON_DATASET_STD_DEV < exchange_rate.metadata.standard_deviation);
```
THLO, line 382: The standard deviation is not necessarily larger, right? If the wrong rates are completely off, they will be discarded and do not matter for the calculation of the standard deviation. In this case, the standard deviation remains basically unchanged. It is only larger if the rates are off but not enough to be discarded. I think this point needs to be made clearer.

THLO (approved):

PR #229, opened 2023-06-23
fix(stddev): Update standard deviation to account for overflow

dfinity-ryancroote:


src/xrc/src/utils.rs
```diff
@@ -232,4 +232,28 @@ pub(crate) mod test {
         );
         assert_eq!(request.timestamp, Some(1234));
     }
+
+    #[test]
+    fn std_dev() {
```
THLO, line 237: Let's add a test that maximizes the standard deviation!
Input: `[u64::MAX, 0]`

src/xrc/src/utils.rs
```diff
@@ -59,16 +59,16 @@ pub(crate) fn standard_deviation(rates: &[u64]) -> u64 {
...
+        .saturating_div(count as u128)) as i128;
     let variance = rates
         .iter()
-        .map(|rate| (((*rate as i64).saturating_sub(mean)).pow(2)) as u64)
-        .sum::<u64>()
-        .saturating_div(count.saturating_sub(1));
+        .map(|rate| ((*rate as i128).saturating_sub(mean)).saturating_pow(2) as u128)
+        .sum::<u128>()
+        .saturating_div(count.saturating_sub(1) as u128);
     (variance as f64).sqrt() as u64
```
THLO, line 72: Is this the only place where we use a floating-point operation? Parsing and displaying floating-point numbers don't count.   :slightly_smiling_face:

I think we can get rid of this operation here. I'll try to address it in a separate PR.

THLO (approved):

src/xrc/src/utils.rs
dfinity-ryancroote: This is the only place if we are not counting the multiplying that occurs when extracting rates.

PR #230, opened 2023-06-27
Removing floating-point square-root operations

THLO:
This PR removes the square-root operations on floating-point numbers, which require a conversion to floating-point numbers and then back.
Instead, an integer square-root function is introduced and used. The function has the following property:
If `y = integer_sqrt(x)`, then `y*y <= x` and `(y+1) * (y+1) > x`.

It is worth noting that the square-root function over real numbers doesn't have this property, so the new function is in some sense more precise.

### DIFF
--- a/src/xrc/src/utils.rs
+++ b/src/xrc/src/utils.rs
@@ -58,7 +58,7 @@ pub(crate) fn median_in_set(values: &[u64]) -> u64 {
 /// value. Since the function is used to compute the standard deviation of 64-bit numbers given
 /// its variance, this limitation is not a concern because the variance is always lower than this
 /// upper bound.
-fn integer_sqrt(number: u128) -> u64 {
+pub(crate) fn integer_sqrt(number: u128) -> u64 {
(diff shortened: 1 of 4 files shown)


src/xrc/src/utils.rs
```diff
@@ -69,7 +96,8 @@ pub(crate) fn standard_deviation(rates: &[u64]) -> u64 {
         .map(|rate| ((*rate as i128).saturating_sub(mean)).saturating_pow(2) as u128)
         .sum::<u128>()
         .saturating_div(count.saturating_sub(1) as u128);
-    (variance as f64).sqrt() as u64
+    println!("Variance: {}", variance);
```
dfinity-ryancroote, line 99: Can this be removed?

THLO: Oops. Yes, it can. I'll remove it. Thanks!

THLO: Done.

PR #231, opened 2023-07-05
`chore`: Improve workflows for release

dfinity-ryancroote:
This PR updates the GitHub workflows to allow to create a release via a tag.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,7 +8,6 @@ on:
   push:
     tags:
       - "*"
-  pull_request:
 
 env:
   DFX_VERSION: "0.14.1"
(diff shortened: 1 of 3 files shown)


pernicat (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,35 @@
...
+
+name: "release"
+
+on:
+  push:
+    tags:
+      - "*"
+
+env:
+  DFX_VERSION: "0.14.1"
```
THLO, line 13: There are some issues with this version, right?
Would it make sense to wait for the 0.14.2 release or is this not necessary?

THLO (approved):

.github/workflows/release.yml
dfinity-ryancroote: It's not necessary. I believe 0.14.2 has been released. I can bump the version in a separate PR.

PR #232, opened 2023-07-06
Update cycle pricing

THLO:
The subnet `uzr34` was recently updated and now charges the [lower prices](https://internetcomputer.org/docs/current/developer-docs/gas-cost) for HTTPS outcalls.

This PR adapts the pricing of the exchange rate canister accordingly, reducing all prices by a factor of 10.


yotamhc (approved):
LGTM, I verified that it aligns with the actual new cost of outcalls on a 34-nodes subnet (now that the cost is quadratic with the subnet size). It still makes sense. 240M cycles is the cost for one outcall with about 1KB request size and 2KB response size.

dfinity-ryancroote (approved):

PR #257, opened 2024-02-12
fix: Remove access by index when inverting and dividing rates

dfinity-ryancroote:
If no rates are retained while instantiating a `QueriedExchangeRate`, the inner `rates` field can become empty. This causes a panic when inverting and dividing rates. This PR patches this behavior to allow the logic to reach the inevitable `validate` calls the occur in each `handle_*_pair` function in `api.rs`.

This is not a major bug as we have implemented `Drop` for our various guards, but it is still a good idea to remove panics.

### DIFF
--- a/src/xrc/src/api.rs
+++ b/src/xrc/src/api.rs
@@ -95,8 +95,6 @@ impl CallExchanges for CallExchangesImpl {
             return Err(CallExchangeError::NoRatesFound);
         }
 
-        rates.sort();
-
         Ok(QueriedExchangeRate::new(
             asset.clone(),
             usdt_asset(),
--- a/src/xrc/src/lib.rs
+++ b/src/xrc/src/lib.rs
@@ -394,7 +394,7 @@ impl std::ops::Mul for QueriedExchangeRate {
 
         // At this stage, the rates are still scaled by `self.decimals` and `other_rate.decimals`.
         // The median rate is used to determine if and how the scaling is adjusted.
-        let median_rate = all_rates[all_rates.len() / 2];
+        let median_rate = all_rates.get(all_rates.len() / 2).cloned().unwrap_or(0u128);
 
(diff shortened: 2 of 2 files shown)


src/xrc/src/api.rs
```diff
@@ -95,8 +95,6 @@ impl CallExchanges for CallExchangesImpl {
             return Err(CallExchangeError::NoRatesFound);
         }
 
-        rates.sort();
```
THLO, line 98: How about sorting the rates here if `!rates.is_empty()`?

src/xrc/src/lib.rs
```diff
@@ -1505,4 +1505,48 @@ mod test {
...
+
+    #[test]
+    /// The function verifies that the [QueriedExchangeRate] struct can be inverted correctly despite
+    /// having been provided an empty rates array.
+    fn inverting_queried_exchange_rate_with_no_rates() {
+        let rate = QueriedExchangeRate::new(btc_asset(), usd_asset(), 0, &[], 0, 0, None);
+
+        assert!(matches!(
+            rate.inverted().validate(),
+            Err(ExchangeRateError::Other(_))
```
THLO, line 1517: What is the actual error?

THLO (approved):

src/xrc/src/api.rs
THLO: I'm a bit confused. If the rates are no longer sorted here, where are they sorted then?

dfinity-ryancroote: The rates are sorted in `QueriedExchangeRate::new` [here](https://github.com/dfinity/exchange-rate-canister/blob/main/src/xrc/src/lib.rs#L533).

src/xrc/src/lib.rs
dfinity-ryancroote: Updated the test to check code and message.

PR #262, opened 2024-03-25
feat: Add flag to support sending cycles on an application subnet.

dfinity-ryancroote:
This PR adds a flag to the compilation process to enable supporting running the XRC on application subnets.


src/xrc/src/exchanges.rs
```diff
@@ -120,6 +120,18 @@ macro_rules! exchanges {
...
                 utils::is_ipv4_support_available() || self.supports_ipv6()
             }
+
+            /// This method returns the number of cycles expected to be sent when
+            /// calling an exchange. As most exchanges are relatively equal,
+            /// we use a static value when on application subnet that includes
+            /// a slight buffer.
+            pub fn cycles(&self) -> u128 {
+                if cfg!(feature = "application-subnet") {
+                    8_000_000
```
THLO, line 130: This value has to be at least the maximum required over all exchanges, right?

THLO (approved):

src/xrc/src/exchanges.rs
dfinity-ryancroote: Good point. I'll update the comments.

PR #263, opened 2024-03-26
fix: Enable beta release by upping the required cycles and lowering the minimum forexes if IPv4 support is not available.

dfinity-ryancroote:
This PR does the following:
1. Raises the required number of cycles needed to submit HTTPS requests on a 13-node subnet.
2. Lowers the minimum number of forexes required to return a forex rate if IPv4 is not available.
3. Fixes a bug in the build-wasm script if no features are enabled then we should just compile the wasm.

### DIFF
--- a/src/monitor-canister/src/periodic.rs
+++ b/src/monitor-canister/src/periodic.rs
@@ -105,7 +105,7 @@ fn make_get_exchange_rate_requests(timestamp: u64) -> Vec<GetExchangeRateRequest
                 class: AssetClass::Cryptocurrency,
             },
             Asset {
-                symbol: "BTT".to_string(),
+                symbol: "USDT".to_string(),
                 class: AssetClass::Cryptocurrency,
             },
             timestamp,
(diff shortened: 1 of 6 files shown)


nategurian (approved):

src/monitor-canister/src/periodic.rs
```diff
@@ -105,7 +105,7 @@ fn make_get_exchange_rate_requests(timestamp: u64) -> Vec<GetExchangeRateRequest
                 class: AssetClass::Cryptocurrency,
             },
             Asset {
-                symbol: "BTT".to_string(),
+                symbol: "USDT".to_string(),
```
THLO, line 108: Why change the symbol?
Note that BTT is used for a reason: The BTC/BTT rate is a good test case because the BTT token has a very small value.
Note that the comment above has not been updated yet (if the symbol should indeed be changed).

dfinity-ryancroote: @THLO That was a mistake to leave those changes in. Thank you for catching them. I have reverted them and the beta XRC is handling them properly.

THLO (approved):

PR #265, opened 2024-05-17
chore: add Subnet Rental Canister to PRIVILEGED_CANISTER_IDS

mraszyk:
This PR adds the Subnet Rental Canister deployed to the NNS subnet (NNS [proposal](https://dashboard.internetcomputer.org/proposal/129683)) to the list of privileged canisters (`PRIVILEGED_CANISTER_IDS`). The Subnet Rental Canister makes at most one request to the Exchange Rate Canister per day querying the ICP/CXDR exchange rate at midnight.

### DIFF
--- a/src/xrc/src/lib.rs
+++ b/src/xrc/src/lib.rs
@@ -71,11 +71,13 @@ pub const XRC_MINIMUM_FEE_COST: u64 = 1_000_000;
 /// The maximum relative difference between accepted rates is 20%.
 pub const MAX_RELATIVE_DIFFERENCE_DIVISOR: u64 = 5;
 
-const PRIVILEGED_CANISTER_IDS: [Principal; 3] = [
+const PRIVILEGED_CANISTER_IDS: [Principal; 4] = [
     // CMC: rkp4c-7iaaa-aaaaa-aaaca-cai
     Principal::from_slice(&[0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x04, 0x01, 0x01]),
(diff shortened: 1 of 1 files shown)


THLO:
Did you consider getting the rate from the CMC instead (fetching the current rate around midnight)?

mraszyk:
> Did you consider getting the rate from the CMC instead (fetching the current rate around midnight)?

What would be the advantage of that as we wouldn't be able to specify a timestamp?

mraszyk:
> The advantage would be that the XRC wouldn't have to provide preferential treatment to another canister.

Preferential treatment is not strictly needed (if that's an issue) - we already topped up the SRC with enough cycles for the next couple of years to come (if the fee for requesting the rate doesn't go up significantly).

mraszyk:
@THLO Feel free to close the MR if you don't like to give the Subnet Rental Canister preferential treatment.

THLO:
> @THLO Feel free to close the MR if you don't like to give the Subnet Rental Canister preferential treatment.

Note that I'm not completely against it, I just wanted to raise this concern. We can discuss it in person if you like.
However, if the canister has enough cycles, then it sounds to me like no preferential treatment is required.

Pro tip: If you fetch the rate _after_ the CMC, you'll get the rate from the cache for only 20M cycles, which is roughly 0.01 USD per year at one call per day.   :slightly_smiling_face:

mraszyk:
> if the canister has enough cycles

it has 19T cycles at the moment and makes at most one call per day so we should be good

PR #270, opened 2024-08-30
docs: Explain exchange rate terminology.

daniel-wong-dfinity-org:
In particular, what does "X/Y" mean? What are "base" and "quote" assets?

In short in "X/Y", X is the "base" asset, and Y is the "quote" asset, and it tells you the price of X in terms of Y.

At least for me, this is really not self-explanatory.

### DIFF
--- a/src/ic-xrc-types/src/lib.rs
+++ b/src/ic-xrc-types/src/lib.rs
@@ -23,8 +23,9 @@ pub struct Asset {
 /// Suppose BTC/USDT is 30_000. This means that you have to give 30_000 USDT in
 /// order to receive 1 BTC. For people with a non-economics background, this is
 /// completely backwards: you would say that the "rate" in this example is
-/// "30_000 USDT per BTC", which would be written as "30_000 USDT/BTC". In this
-/// example, BTC is the "base" asset, and USDT is the "quote" asset.
+/// "30_000 USDT per BTC", which would be written as "30_000 USDT/BTC".
+/// Economists do NOT read "BTC/USD" as "BTC per USD". In this example, BTC is
+/// the "base" asset, and USDT is the "quote" asset.
 ///
 /// Translation table for non-economists:
 ///
--- a/src/ic-xrc-types/src/lib.rs
+++ b/src/ic-xrc-types/src/lib.rs
@@ -20,12 +20,13 @@ pub struct Asset {
 
 /// The type the user sends when requesting a rate.
 ///
-/// For definitions of "base", "quoted", etc, refer to https://en.wikipedia.org/wiki/Currency_pair
+/// For definitions of "base", "quote", etc, the reader is referred to
+/// https://en.wikipedia.org/wiki/Currency_pair.
 #[derive(CandidType, Clone, Debug, Deserialize)]
 pub struct GetExchangeRateRequest {
     /// The base asset.
     pub base_asset: Asset,
-    /// The quoted asset.
+    /// The quote asset.
     pub quote_asset: Asset,
     /// An optional parameter used to find a rate at a specific time.
     pub timestamp: Option<u64>,
@@ -56,7 +57,7 @@ pub struct ExchangeRateMetadata {
 pub struct ExchangeRate {
     /// The base asset.
     pub base_asset: Asset,
-    /// The quoted asset.
+    /// The quote asset.
     pub quote_asset: Asset,
     /// The timestamp associated with the returned rate.
     pub timestamp: u64,
--- a/src/ic-xrc-types/src/lib.rs
+++ b/src/ic-xrc-types/src/lib.rs
@@ -24,9 +24,11 @@ pub struct Asset {
 /// https://en.wikipedia.org/wiki/Currency_pair.
 #[derive(CandidType, Clone, Debug, Deserialize)]
 pub struct GetExchangeRateRequest {
-    /// The base asset.
+    /// The base asset, i.e., the first asset in a currency pair. For example,
+    /// ICP is the base asset in the currency pair ICP/USD.
     pub base_asset: Asset,
-    /// The quote asset.
+    /// The quote asset, i.e., the second asset in a currency pair. For example,
+    /// USD is the quote asset in the currency pair ICP/USD.
     pub quote_asset: Asset,
     /// An optional parameter used to find a rate at a specific time.
     pub timestamp: Option<u64>,
--- a/src/ic-xrc-types/src/lib.rs
+++ b/src/ic-xrc-types/src/lib.rs
@@ -20,23 +20,12 @@ pub struct Asset {
 
 /// The type the user sends when requesting a rate.
 ///
-/// Suppose BTC/USDT is 30_000. This means that you have to give 30_000 USDT in
-/// order to receive 1 BTC. For people with a non-economics background, this is
-/// completely backwards: you would say that the "rate" in this example is
-/// "30_000 USDT per BTC", which would be written as "30_000 USDT/BTC".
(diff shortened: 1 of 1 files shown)


src/ic-xrc-types/src/lib.rs
```diff
@@ -19,13 +19,24 @@ pub struct Asset {
...
+/// completely backwards: you would say that the "rate" in this example is
+/// "30_000 USDT per BTC", which would be written as "30_000 USDT/BTC".
+/// Economists do NOT read "BTC/USD" as "BTC per USD". In this example, BTC is
+/// the "base" asset, and USDT is the "quote" asset.
+///
+/// Translation table for non-economists:
+///
+///             X/Y = Ys per X
+///      base asset = resource that you want to receive
+///     quote asset = resource that you are offerring
```
THLO, line 34: I like the intent but it's too verbose for my liking.
How about we just refer to the definition of [currency pair](https://en.wikipedia.org/wiki/Currency_pair)?
I think it's more important to point out that `X` is the base asset and `Y` is the quote asset in the expression `X/Y` and explain "base asset" and "quote asset". Of course, adding a small example may help.

src/ic-xrc-types/src/lib.rs
```diff
@@ -19,13 +19,24 @@ pub struct Asset {
...
 #[derive(CandidType, Clone, Debug, Deserialize)]
 pub struct GetExchangeRateRequest {
-    /// The asset to be used as the resulting asset. For example, using
-    /// ICP/USD, ICP would be the base asset.
+    /// The resource being sought.
     pub base_asset: Asset,
-    /// The asset to be used as the starting asset. For example, using
-    /// ICP/USD, USD would be the quote asset.
+    /// The resource being offerred.
```
THLO, line 39: I don't think we should use words like "sought" and "offered".
I think if we define base asset and quote asset above, it should be clear here.

daniel-wong-dfinity-org: Done.

# Tangent.

Today, I am under a new understanding (which might also be incorrect): X/Y does **not** indicate which direction you are going. You could just as well be going from X to Y, or from Y to X. The direction would be indicated elsewhere.

I also learned (thanks to the article that you shared) that X is the "more important" one, but there is no globally agreed upon ranking of the world's currencies.

src/ic-xrc-types/src/lib.rs
```diff
@@ -19,13 +19,13 @@ pub struct Asset {
 }
 
 /// The type the user sends when requesting a rate.
+///
+/// For definitions of "base", "quoted", etc, refer to https://en.wikipedia.org/wiki/Currency_pair
```
THLO, line 23:
```suggestion
/// For definitions of "base", "quoted", etc, refer to https://en.wikipedia.org/wiki/Currency_pair.
```

src/ic-xrc-types/src/lib.rs
```diff
@@ -19,13 +19,13 @@ pub struct Asset {
...
 #[derive(CandidType, Clone, Debug, Deserialize)]
 pub struct GetExchangeRateRequest {
-    /// The asset to be used as the resulting asset. For example, using
-    /// ICP/USD, ICP would be the base asset.
+    /// The base asset.
     pub base_asset: Asset,
-    /// The asset to be used as the starting asset. For example, using
-    /// ICP/USD, USD would be the quote asset.
+    /// The quoted asset.
```
THLO, line 28: While not explicitly mentioned in the currency-pair wiki page, it is usually called "quote asset" (and not "quoted asset"). Google Trends agrees with this.   :slightly_smiling_face:

src/ic-xrc-types/src/lib.rs
```diff
@@ -56,7 +56,7 @@ pub struct ExchangeRateMetadata {
 pub struct ExchangeRate {
     /// The base asset.
     pub base_asset: Asset,
-    /// The quote asset.
+    /// The quoted asset.
```
THLO, line 59: Same here.

dskloetd: Regardless of "quote" vs. "quoted", putting a comment `// The foo` on a field called `foo`, doesn't add any value IMO. Can we just remove this?

src/ic-xrc-types/src/lib.rs
daniel-wong-dfinity-org: Done.

src/ic-xrc-types/src/lib.rs
daniel-wong-dfinity-org: > doesn't add any value

👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍

FWIW, it was already like that.

I'd prefer to just delete it, but I don't own this, and deleting would be outside the original intent of this PR -> I do as Thomas says 🙇

src/ic-xrc-types/src/lib.rs
daniel-wong-dfinity-org: Done.

src/ic-xrc-types/src/lib.rs
THLO: :laughing: 
I agree with @dskloetd (and you). Feel free to delete these comment lines as part of this PR!

src/ic-xrc-types/src/lib.rs
```diff
@@ -19,13 +19,14 @@ pub struct Asset {
...
 
 /// The type the user sends when requesting a rate.
+///
+/// For definitions of "base", "quote", etc, the reader is referred to
+/// https://en.wikipedia.org/wiki/Currency_pair.
 #[derive(CandidType, Clone, Debug, Deserialize)]
 pub struct GetExchangeRateRequest {
-    /// The asset to be used as the resulting asset. For example, using
-    /// ICP/USD, ICP would be the base asset.
+    /// The base asset.
```
dskloetd, line 27: I think the comment that was here before had some value but now it's a comment that adds no value.

Actually, I think the example was more useful than a link to some long definition.

dskloetd: As before, it's up to @THLO to agree or not.

THLO, line 27:
```suggestion
    /// The base asset, i.e., the first asset in a currency pair. For example, ICP is the base asset in the currency pair ICP/USD.
```

src/ic-xrc-types/src/lib.rs
```diff
@@ -19,13 +19,14 @@ pub struct Asset {
...
     pub base_asset: Asset,
-    /// The asset to be used as the starting asset. For example, using
-    /// ICP/USD, USD would be the quote asset.
+    /// The quote asset.
```
THLO, line 29:
```suggestion
    /// The quote asset, i.e., the second asset in a currency pair. For example, USD is the quote asset in the currency pair ICP/USD.
```

src/ic-xrc-types/src/lib.rs
daniel-wong-dfinity-org: Thomas made a suggestion. I'm going to take it.

daniel-wong-dfinity-org: Done.

src/ic-xrc-types/src/lib.rs
daniel-wong-dfinity-org: Done.

THLO (approved):

PR #290, opened 2026-01-30
chore: DEFI-2599: Bump rust

mbjorkqvist:
Bump the rust version from 1.75.0 to 1.90.0. Also bump other dependencies that were required to make the build succeed and keep the tests passing, and finally bump the rest of the dependencies as long as no new warnings or test failures were introduced.


gregorydemay (approved):
Thanks @mbjorkqvist for starting this effort!

PR #300, opened 2026-05-05
feat: Add NNS Governance to the list of privileged canisters

jasonz-dfinity:
## Summary
- Adds NNS Governance (`rrkah-fqaaa-aaaaa-aaaaq-cai`) to `PRIVILEGED_CANISTER_IDS`, exempting it from cycle costs when calling the XRC.
- Adds a matching `nns_governance_id_is_correct` test mirroring the existing CMC / NNS-dapp tests.

## Motivation
Phase 2 of Mission 70 moves the maturity modulation computation into NNS Governance, which will pull ICP rate history directly from the XRC instead of relying on CMC. Treating Governance as a system caller (the same status CMC and the NNS dapp already have) avoids the per-call cycle charge on this internal NNS-to-NNS path.

### DIFF
--- a/src/xrc/src/environment.rs
+++ b/src/xrc/src/environment.rs
@@ -127,7 +127,7 @@ pub(crate) mod test {
     impl Default for TestEnvironment {
         fn default() -> Self {
             Self {
-                caller: Principal::from_text("rrkah-fqaaa-aaaaa-aaaaq-cai")
+                caller: Principal::from_text("q4eej-kyaaa-aaaaa-aaaha-cai")
                     .expect("Failed to create test principal"),
                 cycles_available: Default::default(),
                 cycles_accepted: Default::default(),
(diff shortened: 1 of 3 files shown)


mbjorkqvist (approved):
LGTM @jasonz-dfinity! The clippy and `e2e-tests` failures should be fixed if you pull in the latest changes from `main`.

mbjorkqvist:
Some tests seem to be failing since in ones where we verify the behavior of a caller that is not in the allowlist, [the caller canister ID used is the one of the governance canister](https://github.com/dfinity/exchange-rate-canister/blob/c6bdf63cc711537dbbe1284bfee9e6c02d32e487/src/xrc/src/environment.rs#L130). If this is changed to something else (e.g., to [`q4eej-kyaaa-aaaaa-aaaha-cai`](https://dashboard.internetcomputer.org/canister/q4eej-kyaaa-aaaaa-aaaha-cai) of the ICP ledger archive canister), the tests should pass.

PR #302, opened 2026-05-06
fix: replace ondrej/nginx PPA with OpenResty for e2e tests

mbjorkqvist:
The ppa:ondrej/nginx PPA started returning 403 errors, breaking the base Docker image build. libnginx-mod-http-lua is not available in Ubuntu 22.04's standard repos, so switch to OpenResty which bundles nginx with Lua and has its own stable apt repository.

Also add set -euo pipefail to scripts/e2e-tests so a base image build failure surfaces immediately rather than producing confusing downstream errors.

### DIFF
--- a/scripts/e2e-tests
+++ b/scripts/e2e-tests
@@ -2,6 +2,8 @@
 
 # vim: ft=bash
 
+set -euo pipefail
+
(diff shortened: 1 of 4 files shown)


gregorydemay (approved):
Thanks @mbjorkqvist !

PR #301, opened 2026-05-06
chore: Change to using non-deprecated rand functions

mbjorkqvist:
Version `0.9.0` of `rand` deprecated functions such as `rand::thread_rng` and `rand::Rng::gen_range`. Commit [873fb7c5b20270fa413673f9c81602bcbab407e6](https://github.com/dfinity/exchange-rate-canister/commit/873fb7c5b20270fa413673f9c81602bcbab407e6) that bumped the `rand` dependency from `0.8.5` to `0.9.3` did not change to the new functions, and now `clippy` is complaining. To keep `clippy` happy, this PR proposes to change to the new, non-deprecated functions.

PR #303, opened 2026-05-06
ci: make checks-pass actually fail when dependencies fail or skip

basvandijk:
## Problem

The branch protection rule for `main` requires the `checks-pass` job to pass before a PR can merge. However, [PR #301](https://github.com/dfinity/exchange-rate-canister/pull/301) was able to merge while `checks-pass` did not pass.

## Cause

Without `if: always()`, a job inherits the implicit `success()` condition on its `needs`. So when any required job **fails, is cancelled, or is skipped**, `checks-pass` is itself **skipped** rather than failed.

GitHub branch protection treats a skipped required check as passing, which allowed PR #301 to be merged despite the gate.

## Fix

- Add `if: always()` to `checks-pass` so it runs unconditionally.
- Add an explicit step that fails the job if any dependency's result is `failure`, `cancelled`, or `skipped`.

This way `checks-pass` reports a real success/failure status that branch protection can rely on.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -154,10 +154,20 @@ jobs:
         env:
           SHELLCHECK_OPTS: -e SC1090 -e SC2119 -e SC1091
 
-  # Final stage that will fail if the any of the entries in `needs` fails.
+  # Final stage that will fail if any of the entries in `needs` fails or is skipped.
+  # NOTE: `if: always()` is required so this job runs even when a dependency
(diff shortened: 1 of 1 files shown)


mbjorkqvist (approved):
Thanks for the investigation and the fix, @basvandijk! Once [the fix for e2e-tests](https://github.com/dfinity/exchange-rate-canister/pull/302) is merged, we should be able to merge this one also.

PR #305, opened 2026-05-11
fix: Make docker build on x86

Dfinity-Bjoern:
Fix build on Apple Silicon by running x86 image (via macOS Rosetta).


gregorydemay (approved):
Thanks @Dfinity-Bjoern !

PR #306, opened 2026-05-12
fix: DEFI-2817: Replace DAI with USDS due to MakerDAO to Sky Protocol rebranding

mbjorkqvist:
MakerDAO rebranded DAI to USDS. Update the XRC to use the new symbol across all exchange integrations:

 - Rename the `DAI` constant to `USDS` (value "DAI" → "USDS") in `lib.rs`
 - Update `STABLECOIN_BASES` in `api.rs` to use `USDS`
 - Update stablecoin pair configurations for all exchanges
 - Update all tests and doc comments

All exchange APIs were verified live against the USDS symbol. Two issues were found and fixed beyond the straightforward rename:

- KuCoin: The old pair (USDT, DAI) (i.e. USDT-DAI, inverted) no longer exists. KuCoin now lists the pair as USDS-USDT (USDS as base). The pair was updated to (USDS, USDT), which also means the rate is no longer inverted - the invert = `base_symbol == USDT` logic in `get_stablecoin_rate` handles this correctly.
- Crypto.com: Crypto.com does support USDS following the rebrand (DAI was retired on their platform on April 10, 2026), but their only USDS instrument is USDS_USD (priced against USD, not USDT). Since the XRC's stablecoin rate framework exclusively handles USDT-quoted pairs, this pair is incompatible without significant rework. Crypto.com's USDS pair has therefore been removed, leaving it with [(USDT, USDC)]. It still contributes a stablecoin rate via USDC/USDT.

The remaining five exchanges (OKX, GateIo, Mexc, Poloniex, KuCoin) all return valid USDS/USDT candle data.

Renaming DAI to USDS does not break anything wrt. e.g., historical rates, since the `ExchangeRateCache` (crypto/stablecoin rates) is in-memory only and anyway wiped on every upgrade.


src/xrc-tests/src/tests/basic_exchange_rates.rs
```diff
@@ -53,10 +53,10 @@ use crate::{
 /// 1. The XRC retrieves the BTC/USDT rates from the mock exchange responses (request 1 responses).
 ///     1. BTC/USDT rates: [ 41960000000, 42030000000, 42640000000, 44000000000, 44250000000, 44833000000, 44930000000, 46022000000, 46101000000]
 /// 2. The XRC retrieves the stablecoin rates from the mock exchanges.
-///     1.  DAI:  [ 950000000, 990000000, 990000000, 1000000000, 1020000000, 1030927835 ]
-///     2. USDC: [ 950000000, 990000000, 990000000, 1000000000, 1020000000, 1030927835 ]
+///     1.  USDS:  [ 950000000, 970000000, 990000000, 1000000000, 1020000000 ]
```
mbjorkqvist, line 56: The values here changed because:

  1. KuCoin's pair direction flipped. Old `(USDT, DAI)` had `base = USDT`, which triggered `invert = true` in `api.rs:803`, turning the mock's `0.97` into `10¹⁸ / 970_000_000 = 1_030_927_835` (bottom of the list). New `(USDS,
  USDT)` has `base = USDS`, so `invert = false` and the rate stays at `0.97 = 970_000_000 (now second)`. The pair had to flip because KuCoin's real API only exposes the `USDS-USDT` symbol - `USDT-USDS` returns `"Unsupported trading pair"`.
  2. CryptoCom's `USDS` row was removed entirely. CryptoCom doesn't list a `USDS/USDT` market (their `USDS` instrument is `USDS_USD`), so the old `990_000_000` contribution disappeared.

So the second slot used to be one of the two CryptoCom/Poloniex 990s; now it's KuCoin's freshly-uninverted 970.

src/xrc/src/exchanges.rs
```diff
@@ -499,7 +499,7 @@ impl IsExchange for CryptoCom {
     }
 
     fn supported_stablecoin_pairs(&self) -> &[(&str, &str)] {
-        &[(DAI, USDT), (USDT, USDC)]
+        &[(USDT, USDC)]
```
mbjorkqvist, line 494: CryptoCom doesn't list a `USDS/USDT` market (their `USDS` instrument is `USDS_USD`), so the only supported stablecoin pair for them is `(USDT, USDC)`.

src/xrc/src/exchanges.rs
```diff
@@ -290,7 +290,7 @@ impl IsExchange for KuCoin {
     }
 
     fn supported_stablecoin_pairs(&self) -> &[(&str, &str)] {
-        &[(USDC, USDT), (USDT, DAI)]
+        &[(USDC, USDT), (USDS, USDT)]
```
mbjorkqvist: The supported pair at KuCoin is now `(USDS, USDT)`, compared to `(USDT, DAI)` before, so `base = USDT`, requires `invert = true` in `api.rs:803`.

src/xrc/src/exchanges.rs
gregorydemay: I looked at all the overrides of the `IsExchange::supported_stablecoin_pairs` and I think we actually have more USDS sources. Claude found the following:

1. Bitget USDS/USDT and Digifinex USDS/USDT both exist and serve klines -> remove overrides?
2. GateIo actually serves USDC/USDT -> remove override?
3. KuCoin overrides to change the order of the pairs: `(USDC, USDT), (USDS, USDT)` instead of `(USDS, USDT), (USDC, USDT)` by default -> remove override

Also fine as a separate PR.

src/xrc-tests/src/tests/basic_exchange_rates.rs
gregorydemay: Thanks for the explanation!

gregorydemay (approved):
I think we could have more USDS sources but that can also be done in a follow-up PR.

src/xrc/src/exchanges.rs
mbjorkqvist: Good catch, thanks, I didn't think of looking into that, but removing the overrides and retrieving more pairs from more exchanges sounds like a good idea! Done in https://github.com/dfinity/exchange-rate-canister/pull/306/commits/bbb19b72f58e546136c48f40eefad477c58aa66b.

PR #307, opened 2026-05-12
fix: DEFI-2809: Correct MEXC v3 API response parsing

mbjorkqvist:
When MEXC was migrated from API v2 to v3 (the URL was updated to `api.mexc.com/api/v3/klines`), the response parser was not updated to match. The v2 API wrapped results in a JSON object:

```
{"code": 200, "data": [[open_time, open, high, low, close, vol, close_time, quote_vol]]}
```

The v3 API returns a flat array instead:

```
[[open_time, open, high, low, close, vol, close_time, quote_vol]]
```

The `MexcResponse` struct still expected the old data-wrapped format, so every request to MEXC has been failing to parse since the v3 migration. Because parse failures trap the HTTP transform, MEXC has been silently excluded from all rate queries.


src/xrc/src/exchanges.rs
```diff
@@ -380,11 +380,8 @@ impl IsExchange for GateIo {
 }
 
 /// MEXC
-#[derive(Deserialize)]
 #[allow(clippy::type_complexity)]
-struct MexcResponse {
-    data: Vec<(u64, String, String, String, String, String, u64, String)>,
-}
+type MexcResponse = Vec<(u64, String, String, String, String, String, u64, String)>;
```
gregorydemay, line 398: nit: It would help a lot to have some Rust doc here to understand what the different fields represent (see https://www.mexc.com/api-docs/spot-v3/market-data-endpoints/klinecandlestick-data)


Index | Description
-- | --
0 | Open time
1 | Open
2 | High
3 | Low
4 | Close
5 | Volume
6 | Close time
7 | Quote asset volume

gregorydemay (approved):
Thanks @mbjorkqvist !

src/xrc/src/exchanges.rs
mbjorkqvist: Done!

PR #308, opened 2026-05-13
feat: DEFI-2810: Add labeled metrics infrastructure and per-forex observability

mbjorkqvist:
This is the first of three PRs implementing [DEFI-2810](https://dfinity.atlassian.net/browse/DEFI-2810). It introduces the per-source observability we were missing during the MEXC outage and the DAI→USDS migration - incidents where individual exchange/forex sources were silently failing while the aggregate `xrc_errors_returned` counter looked normal.

## What this PR ships

Three new Prometheus series on `/metrics`:

- **`xrc_forex_fetch_total{forex, outcome}`** — counter. Outcomes: `success`, `http_error`, `candid_error`, `empty_map`. Distinguishes "we couldn't reach the upstream" from "the upstream returned HTTP 200 with no rates."
- **`xrc_forex_last_success_seconds{forex}`** — gauge. Wall-clock seconds when each source most recently produced a non-empty rate map. Seeded to `time()` on canister `init` and `post_upgrade`, so a freshly-deployed canister doesn't trip a staleness alert before its first scheduled run.
- **`xrc_periodic_forex_run_last_seconds`** — heartbeat gauge for the 6h periodic task itself. Set on the `Success` return path of `update_forex_store` regardless of per-source success — orthogonal to the per-source metric above. Catches "the periodic task stopped firing" as distinct from "every individual forex returned nothing."

Plus the shared infrastructure those metrics ride on: a small label-aware extension to `MetricsEncoder`, a `LABELED_COUNTERS` / `LABELED_GAUGES` thread-local backing store accessed via `increment_labeled_counter` / `set_labeled_gauge` helpers, and an `init()` hook plumbed from `#[ic_cdk::init]` that also runs at the end of `post_upgrade` to re-seed ephemeral gauges. The labeled state is **not** persisted across upgrades (Prometheus `rate()`/`increase()` already tolerate counter resets, and we didn't want to expand the stable-memory schema).

## What this PR does **not** ship — coming in follow-up PRs

- **PR 2** — per-exchange (`xrc_exchange_fetch_total{exchange, kind, outcome}` + `_last_success_seconds`) and per-stablecoin (`xrc_stablecoin_symbol_rates_received{symbol}`) metrics. This is the PR that would have caught the MEXC and DAI scenarios directly.
- **PR 3** — `xrc_forex_collector_sources` gauge and a cross-cutting integration test under `xrc-tests` that exercises both PR 1 and PR 2 metrics end-to-end against the docker mock.

The existing aggregate counters (`xrc_stablecoin_errors`, `xrc_forex_asset_errors`, etc.) are left untouched in this PR — they overlap semantically with the new labeled metrics, but external dashboards depend on the current names. Coordinating their removal is a separate follow-up after consumers migrate.

## Behavior change worth flagging for log-based alerts

`CallForexError::Empty` is now a distinct enum variant (previously the empty-map case was stuffed into `CallForexError::Http` with a sentinel message string). The canister log line for that case changes:

- Before: `[xrc] <forex> Failed to retrieve rates from <forex>: Empty rates map`
- After:  `[xrc] <forex> Empty rates map from <forex>`

[DEFI-2810]: https://dfinity.atlassian.net/browse/DEFI-2810?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/xrc/src/api/metrics.rs
+++ b/src/xrc/src/api/metrics.rs
@@ -147,6 +147,11 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
         "xrc_forex_last_success_seconds",
         "Unix timestamp (seconds) of the most recent successful fetch per forex source.",
     )?;
+    encode_labeled_gauge_family(
+        w,
+        "xrc_periodic_forex_run_last_seconds",
+        "Unix timestamp (seconds) of the most recent successful periodic forex-update run.",
+    )?;
 
     Ok(())
 }
@@ -278,6 +283,9 @@ impl<W: io::Write> MetricsEncoder<W> {
         help: &str,
     ) -> io::Result<()> {
         self.encode_header(name, help, typ)?;
+        if labels.is_empty() {
+            return writeln!(self.writer, "{} {} {}", name, value, self.now_millis);
+        }
         write!(self.writer, "{}{{", name)?;
         for (i, (k, v)) in labels.iter().enumerate() {
             if i > 0 {
@@ -449,11 +457,12 @@ mod test {
     }
 
     #[test]
-    fn empty_label_set_still_emits_braces() {
+    fn empty_label_set_omits_braces() {
         let out = encode(0, |e| {
             e.encode_counter_with_labels("metric", &[], 1, "h").unwrap();
         });
-        assert!(out.contains("metric{} 1 0"), "got: {out}");
+        assert!(out.contains("metric 1 0"), "got: {out}");
+        assert!(!out.contains("metric{}"), "got: {out}");
     }
 
     #[test]
--- a/src/xrc/src/api/metrics.rs
+++ b/src/xrc/src/api/metrics.rs
@@ -150,7 +150,7 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
     encode_labeled_gauge_family(
         w,
         "xrc_periodic_forex_run_last_seconds",
-        "Unix timestamp (seconds) of the most recent successful periodic forex-update run.",
+        "Unix timestamp (seconds) of the most recent periodic forex-update task run (heartbeat — not contingent on rate-fetch success).",
     )?;
 
     Ok(())
--- a/src/xrc/src/api/metrics.rs
+++ b/src/xrc/src/api/metrics.rs
@@ -1,6 +1,6 @@
 use crate::{
-    rate_limiting, types::HttpResponse, with_cache, with_forex_rate_store, AllocatedBytes,
-    MetricCounter,
+    rate_limiting, types::HttpResponse, with_cache, with_forex_rate_store, with_labeled_counters,
+    with_labeled_gauges, AllocatedBytes, LabelPairs, MetricCounter,
 };
 use ic_cdk::api::time;
 use serde_bytes::ByteBuf;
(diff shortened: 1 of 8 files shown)


src/xrc/src/lib.rs
```diff
@@ -159,6 +160,135 @@ thread_local! {
...
+    LABELED_GAUGES.with(|m| m.borrow_mut().clear());
+}
+
+/// Names of all labeled metrics. Centralised so that the recording site
+/// (e.g. `periodic.rs`), the encoder (`api/metrics.rs`), and the init
+/// seeding can never drift out of sync via typos — a typo at one site
+/// would silently produce a "recorded but never emitted" (or vice versa)
+/// split-brain metric. Every new labeled metric introduced by a follow-up
+/// PR should land its name here first.
+pub(crate) mod metric_names {
```
gregorydemay: Understanding question: why not have the static strings as an enum (similar question for `metric_labels`)? The benefit would be to use then those types in `make_metric_key`, `LabelPairs`, `MetricKey` which would help readability.

src/xrc/src/periodic.rs
```diff
@@ -572,4 +618,172 @@ mod test {
             1680134400
         );
     }
```
gregorydemay, line 620: Maybe that's better for another PR but I think it would help to get an integration test with PocketIC to ensure that the exposed metrics are the expected ones. See this [example](https://github.com/dfinity/dex/blob/a77fa2c1133780a5ac9997644aae06d9ea7039b9/integration_tests/tests/tests.rs#L1441).

src/xrc/src/lib.rs
```diff
@@ -159,6 +160,135 @@ thread_local! {
...
+{
+    LABELED_GAUGES.with(|m| f(&m.borrow()))
+}
+
+/// Initializes ephemeral state that does not survive a canister upgrade.
+/// Called from the `#[ic_cdk::init]` hook and from [`post_upgrade`] after
+/// stable state is restored. Seeds every `*_last_success_seconds` gauge
+/// to the current time so a freshly-deployed canister doesn't immediately
+/// trip a staleness alert.
+pub fn init() {
```
gregorydemay: nit: maybe rename to something like `init_metrics`? Reading `init` in `post_upgrade` is a bit scary, one could think some state is being reset, which is not the case here.

src/xrc/src/api/metrics.rs
```diff
@@ -164,13 +213,30 @@ pub fn heap_memory_size_bytes() -> usize {
...
+    ///   share a header pair;
+    /// - the unlabeled path `debug_assert!`s the name is fresh, since
+    ///   `encode_counter` / `encode_gauge` is meant for single-series
+    ///   metrics and emitting two of them would violate the Prometheus
+    ///   exposition format ("Only one HELP line may exist for any given
+    ///   metric name").
+    ///
+    /// `&'static str` keys (rather than `String`) avoid an allocation per
+    /// first-emit; all callers pass static metric-name constants.
+    headers_emitted: HashSet<&'static str>,
```
gregorydemay: nit: use `BTreeSet` for reproducibility?

gregorydemay (approved):
Thanks @mbjorkqvist ! Some minor comments but otherwise LGTM!

src/xrc/src/lib.rs
mbjorkqvist: Good call — went with the full enum refactor in commit 8f9cd9c. `MetricName`, `LabelKey`, and `Outcome` now live as `strum::IntoStaticStr`-derived enums; the helpers (`make_metric_key`, `increment_labeled_counter`, `set_labeled_gauge`) and the `MetricKey`/`LabelPairs` types take them directly. The closed-set property is in the type system now, and the `match` on `CallForexError` returning `Outcome` is exhaustiveness-checked.

src/xrc/src/periodic.rs
mbjorkqvist: Agreed — PR 3 in this series already plans a cross-cutting integration test under `xrc-tests` that exercises both PR 1's and PR 2's metrics end-to-end against the docker mock (parses `/metrics` from `http_request` and asserts on the labeled series after a forex run). Bundling them into one harness keeps us from landing two near-duplicate test setups. Tracking under [DEFI-2810](https://dfinity.atlassian.net/browse/DEFI-2810).

src/xrc/src/lib.rs
```diff
@@ -159,6 +160,135 @@ thread_local! {
     static CRYPTO_ASSET_RELATED_ERRORS_COUNTER: Cell<usize> = const { Cell::new(0) };
     static FOREX_ASSET_RELATED_ERRORS_COUNTER: Cell<usize> = const { Cell::new(0) };
 
+    /// Per-(metric-name, label-set) labeled metrics. Populated by recording
+    /// sites via [`increment_labeled_counter`] and [`set_labeled_gauge`],
+    /// drained by [`api::metrics::get_metrics`]. Reset on canister upgrade.
```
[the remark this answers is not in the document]
mbjorkqvist: Addressed in f579970 — the doc comment now reads "encoded by `api::metrics::get_metrics`" rather than "drained by".

src/xrc/src/lib.rs
mbjorkqvist: Renamed in commit 9bf9abf — `xrc::init` is now `xrc::init_metrics`. Reading it in `post_upgrade` no longer suggests state is being reset.

src/xrc/src/api/metrics.rs
mbjorkqvist: Swapped in commit a96051d — `headers_emitted` is now a `BTreeSet`.

PR #309, opened 2026-05-19
feat: DEFI-2810: Add per-exchange and per-stablecoin observability

mbjorkqvist:
This is the second of three PRs implementing [DEFI-2810]. Building on PR 1 (#308), which introduced the labelled-metrics infrastructure and per-forex observability, this PR adds the per-exchange and per-stablecoin series — the ones that would have surfaced both the MEXC outage and the DAI→USDS migration directly rather than only via the aggregate `xrc_errors_returned` counter.

## What this PR ships

Three new Prometheus series on `/metrics`:

- **`xrc_exchange_fetch_total{exchange, kind, outcome}`** — counter. The `kind` label distinguishes the two call contexts (`crypto` for USDT-pair lookups, `stablecoin` for stablecoin-symbol lookups). Outcomes: `success`, `http_error`, `candid_error`, `extracted_zero`. The last one captures the MEXC-shaped silent-failure mode — upstream returns 200 with a parseable response whose rate truncates to zero (literal `"0"`, ultra-low-value coin underflowing `rate * RATE_UNIT`, negative/NaN coerced via `as u64`). The canister already silently drops these via `QueriedExchangeRate::new`'s zero-filter and the stablecoin invert-check; this metric promotes the silent defence into observable signal.

- **`xrc_exchange_last_success_seconds{exchange, kind}`** — gauge. Advances only on real `Outcome::Success` — `Ok(0)` doesn't refresh it, so an exchange flipping to all-zero quotes both bumps `extracted_zero` and stops refreshing the staleness gauge. Seeded at `init()` / `post_upgrade` to the current time for every `(exchange, kind)` pair so a freshly-deployed canister doesn't trip a staleness alert before its first request reaches each upstream.

- **`xrc_stablecoin_symbol_rates_received{symbol}`** — counter. Adds the count of *usable* (non-zero) per-exchange rates per `get_stablecoin_rate` run. An empty or all-zero input adds zero, which materialises the series so `rate(...) == 0` alerts have something to evaluate from t=0. This is the metric that directly catches the DAI-rebrand failure mode: when upstreams stop quoting a symbol, `rate(...[1h]) == 0` fires before the aggregate `StablecoinRateTooFewRates` error ever surfaces.

Plus the shared infrastructure those metrics ride on: an `ExchangeCallKind` enum, a delta-aware `add_labeled_counter` sibling of `increment_labeled_counter`, the `record_exchange_outcome` + `record_stablecoin_symbol_rates_received` helpers, and an init-seeding extension. As in PR 1, labelled state is **not** persisted across upgrades — Prometheus `rate()`/`increase()` already tolerates counter resets and we didn't want to expand the stable-memory schema.

## What this PR does **not** ship — coming in PR 3

- `xrc_forex_collector_sources` gauge.
- Cross-cutting integration test under `xrc-tests` exercising PR 1 + PR 2 metrics end-to-end against the docker mock.

The existing aggregate counters (`xrc_stablecoin_errors`, `xrc_errors_returned`, etc.) are left untouched — they overlap semantically with the new labelled metrics, but external dashboards depend on the current names. Coordinating their removal is a separate follow-up after consumers migrate.

## Internal change worth flagging

`async fn call_exchange(exchange, args)` now takes an `ExchangeCallKind` parameter. The original body became `call_exchange_raw`; the new wrapper threads `kind` through `record_exchange_outcome` before forwarding the result. The two call sites in `api.rs` pass `ExchangeCallKind::Crypto` and `ExchangeCallKind::Stablecoin` respectively. No public Candid surface change.

## Semantic divergence for dashboard authors

`xrc_stablecoin_symbol_rates_received{symbol}` will **not** match `ExchangeRateMetadata::base_asset_num_received_rates` for the same query — the metric counts usable (non-zero) per-exchange rates, while the public-API field counts every `Ok` response including zero-valued ones. Both are deliberate: the metric measures data-feed health across runs, the field measures upstream responsiveness for a specific request. The function doc on `record_stablecoin_symbol_rates_received` spells this out.

[DEFI-2810]: https://dfinity.atlassian.net/browse/DEFI-2810


src/xrc/src/api/test.rs
```diff
@@ -2012,3 +2012,97 @@ mod request_contains_invalid_symbols {
         ));
     }
 }
+
+mod stablecoin_symbol_metrics {
+    use super::super::record_stablecoin_symbol_rates_received;
+    use crate::{
+        make_metric_key, reset_labeled_metrics_for_test, with_labeled_counters, LabelKey,
+        MetricName,
+    };
```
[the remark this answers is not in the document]
mbjorkqvist: The code compiles. The module path is `crate::api::test::stablecoin_symbol_metrics` (the `test` module is declared as `pub(crate) mod test;` in api.rs:4), so `super::super` resolves to `crate::api` — which is where `record_stablecoin_symbol_rates_received` is defined (api.rs:888). Verified with `cargo test -p xrc stablecoin_symbol_metrics --no-run`.

src/xrc/src/api/metrics.rs
```diff
@@ -152,6 +162,11 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
         MetricName::PeriodicForexRunLastSeconds,
         "Unix timestamp (seconds) of the most recent periodic forex-update task run (heartbeat — not contingent on rate-fetch success).",
     )?;
+    encode_labeled_counter_family(
+        w,
+        MetricName::StablecoinSymbolRatesReceived,
+        "Count of individual exchange rate samples received per stablecoin symbol; drops to zero when an upstream symbol stops returning data (e.g. after a rebrand).",
```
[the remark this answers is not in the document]
mbjorkqvist: Good catch — already addressed in commit d48cffb ("Fix help text"). The line now reads "…**its rate** drops to zero when an upstream symbol stops returning data…", which matches the `rate(xrc_stablecoin_symbol_rates_received[...]) == 0` framing you suggested.

src/xrc/src/lib.rs
```diff
@@ -183,24 +184,37 @@ pub(crate) fn reset_labeled_metrics_for_test() {
...
 }
 
 /// Declaration order is load-bearing: [`make_metric_key`] sorts label
-/// pairs by `LabelKey`, and keeping `Forex < Outcome` (matching what
-/// `&str` sort of `"forex" < "outcome"` would produce) preserves the
-/// byte-exact `/metrics` output.
+/// pairs by `LabelKey`, and keeping declaration order aligned with the
+/// `&str` sort of the serialized label names (`"exchange" < "forex" <
+/// "kind" < "outcome" < "symbol"`) preserves the byte-exact `/metrics`
+/// output.
```
gregorydemay: nit: if the order is that important, maybe let's have a unit test to ensure that it stays the way it should be?

src/xrc/src/lib.rs
```diff
@@ -213,6 +227,33 @@ pub(crate) enum Outcome {
...
+}
+
+/// Discriminates the two call contexts in which an exchange is queried:
+/// crypto-pair lookups versus stablecoin lookups. Used as the `kind`
+/// label on the per-exchange metric families.
+///
+/// `EnumIter` is derived so callers that need to materialise series
+/// across every variant (e.g. `init_at`'s gauge-seeding loop) iterate
+/// exhaustively at compile time — adding a third variant later won't
+/// silently miss the new kind in those loops.
```
gregorydemay: nit: remove? The advantages of strum::EnumIter are I think clear enough. I usually find Claude too chatty.

src/xrc/src/lib.rs
```diff
@@ -854,6 +931,69 @@ async fn call_exchange(
...
+/// HTTP outcall path.
+///
+/// `CallExchangeError::NoRatesFound` is never produced by
+/// [`call_exchange_raw`] (the HTTP outcall path) — it only arises in
+/// callers that aggregate or post-process per-exchange results (e.g.
+/// the inverted-stablecoin path in `api.rs`). Treating it as a no-op
+/// here is intentional: it isn't a per-call outcome. A `debug_assert!`
+/// on the arm catches a future regression in test builds (where a new
+/// producer would otherwise silently drop observations); release builds
+/// compile the assertion out so the canister never panics on it.
```
gregorydemay: I don't understand this argument and the explanation below in the arm `CallExchangeError::NoRatesFound`: 

1. "`NoRatesFound` is never produced by the only production caller (`call_exchange_raw`); see this function's doc comment." I'm looking at the doc and I don't see such an explanation.
2. Why do we need to know that `CallExchangeError::NoRatesFound` is never produced? That's an implementation detail. The logic would be more robust if we don't need to assume anything.

src/xrc/src/lib.rs
```diff
@@ -213,6 +227,33 @@ pub(crate) enum Outcome {
...
+/// `EnumIter` is derived so callers that need to materialise series
+/// across every variant (e.g. `init_at`'s gauge-seeding loop) iterate
+/// exhaustively at compile time — adding a third variant later won't
+/// silently miss the new kind in those loops.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, strum::IntoStaticStr, strum::EnumIter)]
+pub(crate) enum ExchangeCallKind {
+    #[strum(serialize = "crypto")]
+    Crypto,
+    #[strum(serialize = "stablecoin")]
+    Stablecoin,
```
gregorydemay, line 249: nit: I find the naming of the variants a bit confusing because a stablecoin is obviously also a crypto asset. I don't really have a better naming, maybe just a doc comment would help (mentioning that first one is for `CallExchanges::get_cryptocurrency_usdt_rate` while the second one is for `CallExchanges::get_stablecoin_rates`)

gregorydemay (commented):
Thanks @mbjorkqvist ! Just some minor comments.

src/xrc/src/lib.rs
mbjorkqvist: Good prompt to revisit — when I traced through where the order *actually* matters, it turned out it doesn't, so I went the other way and dropped the invariant rather than locking it in (82aa265).

Walking through it:

- **`MetricName` declaration order**: doesn't affect `/metrics` output at all. `encode_metrics` calls `encode_labeled_counter_family` once per family in a hand-written sequence, and each call filters the BTreeMap by name. The enum's `Ord` only controls the BTreeMap's intra-iteration order, which the filter ignores.
- **`LabelKey` declaration order**: controls the order labels appear *within* a series line (`{exchange=..,kind=..,outcome=..}`). Prometheus treats label sets as unordered, so two scrapes with different intra-line orders are the same series — no dashboard, alert, or `rate()` query cares.
- **Tests**: the only assertions on byte-exact labelled output (metrics.rs:414, :439, :441, :457) test the third-party `MetricsEncoder` directly with hand-written input slices. They don't flow through `make_metric_key`, so `LabelKey` Ord couldn't affect them.

`Ord` still has to stay on both enums because `MetricKey = (MetricName, Vec<(LabelKey, String)>)` is a `BTreeMap` key — but the specific variant order is incidental. Reworded the doc comment to say exactly that.

src/xrc/src/lib.rs
mbjorkqvist: Done in 971423b — agree the derive speaks for itself.

src/xrc/src/lib.rs
mbjorkqvist: Agreed on the ambiguity — "stablecoin" is obviously also crypto. Went with per-variant doc comments pointing at the originating `CallExchanges` method, since that pins down which discriminant maps to which call path without needing to rename either variant (e3923ab).

src/xrc/src/lib.rs
mbjorkqvist: Fair points on both. You're right that (1) the doc didn't actually justify the claim — it just asserted it. And (2) the "this variant can never reach us" reasoning was an implementation-detail assumption masquerading as a contract, with a `debug_assert!` that would silently drop observations in release builds if a future refactor wired up a new producer.

Restructured in 74fedef: added `Outcome::NoRatesFound` and dropped the special case. `record_exchange_outcome` now treats every `CallExchangeError` variant the same way — increment `xrc_exchange_fetch_total` with the appropriate `outcome` label — so the function no longer assumes anything about who its callers are. The match is exhaustive at the type level.

The accompanying test flipped from a `#[cfg(debug_assertions)] #[should_panic]` to a positive assertion that the counter increments and the success gauge does not.

src/xrc/src/lib.rs
```diff
@@ -183,24 +184,35 @@ pub(crate) fn reset_labeled_metrics_for_test() {
...
+    StablecoinSymbolRatesReceived,
 }
 
-/// Declaration order is load-bearing: [`make_metric_key`] sorts label
-/// pairs by `LabelKey`, and keeping `Forex < Outcome` (matching what
-/// `&str` sort of `"forex" < "outcome"` would produce) preserves the
-/// byte-exact `/metrics` output.
+// `Ord` is required because `MetricKey` is used as a `BTreeMap` key;
+// the specific variant order is incidental and not depended on by
+// `/metrics` output or any consumer.
```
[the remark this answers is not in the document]
gregorydemay: I would just remove the whole comment. If `Ord` was not there code would not compile.

gregorydemay (approved):
Thanks @mbjorkqvist !

src/xrc/src/lib.rs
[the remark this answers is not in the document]
mbjorkqvist: Done in fb9cbd0 — agreed, the compiler already enforces it via `MetricKey`'s `BTreeMap` usage. Comment dropped entirely.

PR #310, opened 2026-05-22
test: DEFI-2810: E2e coverage for labeled metrics and doc cleanups

mbjorkqvist:
## Summary

Final canister-side PR for [DEFI-2810](https://dfinity.atlassian.net/browse/DEFI-2810). Builds on #308 (labeled-metrics foundation + per-forex) and #309 (per-exchange + per-stablecoin) by adding end-to-end test coverage for every labeled series introduced across the earlier two PRs, plus small doc cleanups. The corresponding Prometheus alert rules ship separately in [dfinity-ops/k8s#2459](https://github.com/dfinity-ops/k8s/pull/2459).

- **E2e integration test** (`xrc-tests/metrics_endpoint`) — boots the canister against the docker mock, drives a `get_exchange_rate` request with one exchange configured to fail, scrapes `/metrics` via the `http_request` candid endpoint, and asserts every labeled metric family from PRs 1 & 2 is present, including a non-success outcome on the per-exchange counter. Includes a regression guard: each labeled metric name emits exactly one `# HELP` line.
- **Group B `# HELP` annotations** — `xrc_stablecoin_errors`, `xrc_crypto_asset_errors`, and `xrc_forex_asset_errors` now point dashboard authors at the per-source replacements from PRs 1 & 2. Pure docstring change.
- **DEFI-2828 TODO** on the new test's `#[ignore]` attribute, describing the follow-up to replace the implicit "requires Docker mock harness" convention with an explicit gate (reason string / feature flag).

### Note on the dropped collector-sources gauge

An earlier draft of this PR added `xrc_forex_collector_sources{slot}` exposing per-deque-slot source counts. It was dropped during review (commit `9c1497d`) — the collector deque is only mutated on successful source fetches and has no wall-clock expiry, so a sustained outage leaves the gauge reporting stale counts for arbitrarily old days while `== 0` only fires on a cold-start empty deque. The signals operators actually need are already on `/metrics` from #308 / #309 and covered by the new e2e test — and operationalised as alerts in [dfinity-ops/k8s#2459](https://github.com/dfinity-ops/k8s/pull/2459):

| Scenario | Alert |
|---|---|
| All forex data is stale (errors, empty, unreachable, _or_ timer dead) | `time() - min(xrc_forex_last_success_seconds) > 86400` |
| Timer not firing | `time() - xrc_periodic_forex_run_last_seconds > 1800` |
| Any individual source down for an extended period | `time() - xrc_forex_last_success_seconds > 172800` |

## Test plan

- [x] `cargo test -p xrc` — 215 unit tests pass post-revert.
- [x] `./scripts/e2e-tests` — passed at the gauge-included revision (`84aaca7`); the post-revert change to `metrics_endpoint.rs` only removes assertions, which cannot regress a previously passing test.

## Follow-up

[DEFI-2828](https://dfinity.atlassian.net/browse/DEFI-2828) tracks replacing the `#[ignore]` gating in `xrc-tests` with an explicit gate (reason string / feature flag).

[DEFI-2810]: https://dfinity.atlassian.net/browse/DEFI-2810
[DEFI-2828]: https://dfinity.atlassian.net/browse/DEFI-2828

### DIFF
--- a/src/xrc-tests/src/tests/metrics_endpoint.rs
+++ b/src/xrc-tests/src/tests/metrics_endpoint.rs
@@ -20,13 +20,19 @@ use crate::{
 /// * Scrape the `/metrics` endpoint via the canister's `http_request` query.
 ///
 /// Success criteria:
-/// * The labeled series introduced in PR 1 and PR 2 of DEFI-2810 all appear
-///   on `/metrics`, with both success and failure outcomes represented on
-///   the per-exchange counter.
-/// * The `xrc_forex_collector_sources` gauge from this PR is present for
-///   both deque slots.
-/// * Each new metric name emits exactly one `# HELP` line — regression guard
-///   on the labeled-encoder header dedup logic.
+/// * Every labeled metric family is present on `/metrics`:
+///   `xrc_exchange_fetch_total` (with both success and a non-success outcome
+///   represented), `xrc_exchange_last_success_seconds`,
+///   `xrc_forex_fetch_total`, `xrc_forex_last_success_seconds`,
+///   `xrc_periodic_forex_run_last_seconds`,
+///   `xrc_stablecoin_symbol_rates_received`, and
+///   `xrc_forex_collector_sources` (both deque slots).
+/// * Each labeled metric name emits exactly one `# HELP` line — regression
+///   guard on the labeled-encoder header dedup logic.
+// TODO(DEFI-2828): Replace `#[ignore]` with a clearer gate (reason string /
+// feature flag) across the xrc-tests crate so the intent — "requires the
+// Docker mock harness via ./scripts/e2e-tests" — is explicit rather than
+// implicit.
 #[ignore]
 #[test]
 fn metrics_endpoint_exposes_all_labeled_series() {
--- a/src/xrc-tests/src/tests/metrics_endpoint.rs
+++ b/src/xrc-tests/src/tests/metrics_endpoint.rs
@@ -165,12 +165,12 @@ fn metrics_endpoint_exposes_all_labeled_series() {
 
         // Collector-sources gauge — both deque slots always emitted.
         assert!(
-            body.contains(r#"xrc_forex_collector_sources{day_offset="0"}"#),
-            "missing collector_sources day_offset=0\n{body}"
+            body.contains(r#"xrc_forex_collector_sources{slot="newest"}"#),
+            "missing collector_sources slot=newest\n{body}"
         );
         assert!(
-            body.contains(r#"xrc_forex_collector_sources{day_offset="1"}"#),
-            "missing collector_sources day_offset=1\n{body}"
+            body.contains(r#"xrc_forex_collector_sources{slot="previous"}"#),
+            "missing collector_sources slot=previous\n{body}"
         );
 
         // Each labeled metric name should emit exactly one `# HELP` line,
--- a/src/xrc-tests/src/tests/metrics_endpoint.rs
+++ b/src/xrc-tests/src/tests/metrics_endpoint.rs
@@ -1,0 +1,192 @@
+use ic_xrc_types::{Asset, AssetClass, GetExchangeRateRequest, GetExchangeRateResult};
+use maplit::hashmap;
+use xrc::types::{HttpRequest, HttpResponse};
+
+use crate::{
+    container::{run_scenario, Container},
+    mock_responses,
+};
+
+/// Setup:
+/// * Deploy mock FOREX data providers and exchanges, with one exchange
+///   configured to return no body (driving a non-success outcome on the
+///   exchange counter).
+/// * Start replicas and deploy the XRC, configured to use the mock data
+///   sources.
+///
+/// Runbook:
+/// * Drive a `get_exchange_rate` request so the recording sites fire across
+///   the crypto, stablecoin, and forex code paths.
+/// * Scrape the `/metrics` endpoint via the canister's `http_request` query.
+///
+/// Success criteria:
+/// * The labeled series introduced in PR 1 and PR 2 of DEFI-2810 all appear
+///   on `/metrics`, with both success and failure outcomes represented on
+///   the per-exchange counter.
+/// * The `xrc_forex_collector_sources` gauge from this PR is present for
+///   both deque slots.
+/// * Each new metric name emits exactly one `# HELP` line — regression guard
+///   on the labeled-encoder header dedup logic.
+#[ignore]
+#[test]
+fn metrics_endpoint_exposes_all_labeled_series() {
+    let now_seconds = time::OffsetDateTime::now_utc().unix_timestamp() as u64;
+    let timestamp_seconds = now_seconds / 60 * 60;
+
+    let responses = mock_responses::exchanges::build_responses(
+        "ICP".to_string(),
+        timestamp_seconds,
+        |exchange| match exchange {
+            // KuCoin returns no body — drives a non-success outcome on the
+            // crypto fetch path for this exchange.
+            xrc::Exchange::KuCoin(_) => None,
+            xrc::Exchange::Coinbase(_) => Some("3.92"),
+            xrc::Exchange::Okx(_) => Some("3.90"),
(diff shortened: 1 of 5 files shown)


src/xrc-tests/src/tests/metrics_endpoint.rs
```diff
@@ -0,0 +1,192 @@
...
+///   the crypto, stablecoin, and forex code paths.
+/// * Scrape the `/metrics` endpoint via the canister's `http_request` query.
+///
+/// Success criteria:
+/// * The labeled series introduced in PR 1 and PR 2 of DEFI-2810 all appear
+///   on `/metrics`, with both success and failure outcomes represented on
+///   the per-exchange counter.
+/// * The `xrc_forex_collector_sources` gauge from this PR is present for
+///   both deque slots.
+/// * Each new metric name emits exactly one `# HELP` line — regression guard
```
gregorydemay: nit: I would not talk about PRs in the Rust doc. How the work was organized in different PRs should not be relevant here.

src/xrc-tests/src/tests/metrics_endpoint.rs
```diff
@@ -0,0 +1,192 @@
...
+///   on the labeled-encoder header dedup logic.
+#[ignore]
```
gregorydemay, line 35: why ignore?

src/xrc/src/api/metrics.rs
```diff
@@ -167,10 +171,28 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
...
+        currently held in the collector, ordered newest-first \
+        (day_offset=\"0\" is the most recent stored day, \"1\" the day before).";
+    let name: &'static str = MetricName::ForexCollectorSources.into();
+    let counts = with_forex_rate_collector(|c| c.sources_per_day());
+    for (offset, label_value) in ["0", "1"].iter().enumerate() {
+        let value = counts.get(offset).copied().unwrap_or(0) as f64;
+        w.encode_gauge_with_labels(name, &[("day_offset", label_value)], value, HELP)?;
+    }
+    Ok(())
+}
```
gregorydemay: understanding question: why do we want both `day_offset="0"` and `day_offset="1"`? Why not only one or what about the other ones in `counts`?

gregorydemay (commented):
Mostly an understanding question

src/xrc-tests/src/tests/metrics_endpoint.rs
mbjorkqvist: Agreed — reworded in 84aaca7 to describe the metric families directly instead of referring to how the work was split across PRs.

src/xrc-tests/src/tests/metrics_endpoint.rs
mbjorkqvist: Existing convention in xrc-tests — every test in this crate is `#[ignore]`'d so `cargo test` doesn't try to run them without the Docker mock harness; they're driven through `./scripts/e2e-tests` instead. Added a `TODO(DEFI-2828)` next to the attribute (84aaca7) and filed DEFI-2828 to replace the `#[ignore]` gating across the whole crate with a clearer signal (reason string / feature flag) so the intent is explicit rather than implicit.

src/xrc/src/api/metrics.rs
mbjorkqvist: You're right to push on this, and on reflection you were right the first time — I've dropped the gauge from the PR entirely in 9c1497d.

The metric I was defending doesn't carry the signal I was claiming. The collector deque (`ForexRatesCollector.days`) is only mutated by `update()` (`forex.rs:618`), which is only called when a source returns rates (`periodic.rs:217`). There is no wall-clock expiry. So if every source fails for a week, `xrc_forex_collector_sources{slot="newest"}` keeps reporting the count and timestamp of the most recently populated day — a week ago. `== 0` only fires on a cold-start empty deque, not on "everything is broken." The per-slot decomposition didn't rescue this: `slot="previous" == 0` is the *normal* state most of the time, since the second slot only fills during UTC rollover overlap windows. So both your concerns — "why two?" and (implicitly) "is this the right shape at all?" — were valid.

The signals operators actually need are already on `/metrics` from the earlier DEFI-2810 PRs:

| Scenario | Alert |
|---|---|
| All forex data is stale (errors, empty, unreachable, *or* timer dead) | `time() - min(xrc_forex_last_success_seconds) > 86400` |
| Timer not firing (distinguishes from "all sources failing") | `time() - xrc_periodic_forex_run_last_seconds > 1800` |
| Any individual source down for an extended period | `time() - xrc_forex_last_success_seconds > 172800` |
| Sustained errors on a specific source | `rate(xrc_forex_fetch_total{outcome!="success"}[1h]) > 0 unless rate(xrc_forex_fetch_total{outcome="success"}[1h]) > 0` |

These cover the failure modes a collector-sources gauge couldn't catch: error storms (`forex_fetch_total{outcome}`), timer wedging (`periodic_forex_run_last_seconds`), and stale data with a still-populated deque (`forex_last_success_seconds`). Adding a `xrc_forex_collector_sources == 0` series alongside them would have been actively misleading — it looks alertable but only fires on cold start, so anyone adopting it would think they had coverage they didn't have.

What's left in the PR after the revert: the e2e test for the labeled-series surface (95f2ec2) and the doc/`# HELP` cleanups on the Group B aggregates (b13b388) — the test still guards every metric we're keeping. Apologies for the detour.

src/xrc-tests/src/tests/metrics_endpoint.rs
```diff
@@ -0,0 +1,186 @@
...
+                symbol: "ICP".to_string(),
+                class: AssetClass::Cryptocurrency,
+            },
+            quote_asset: Asset {
+                symbol: "EUR".to_string(),
+                class: AssetClass::FiatCurrency,
+            },
+        };
+        let _ = container
+            .call_canister::<_, GetExchangeRateResult>("get_exchange_rate", &request);
```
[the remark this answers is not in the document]
mbjorkqvist: Good catch — applied in 418b1b4. The outer `call_canister` is now `.expect`-ed so a dfx-level or candid-decode failure surfaces with a clear root-cause; the inner `GetExchangeRateResult` is still intentionally discarded (value correctness lives in `basic_exchange_rates.rs` / `misbehavior.rs`). The test comment is updated to reflect that split.

src/xrc-tests/src/tests/metrics_endpoint.rs
```diff
@@ -0,0 +1,186 @@
...
+
+        assert_eq!(response.status_code, 200, "/metrics returned non-200");
+        let body =
+            String::from_utf8(response.body.into_vec()).expect("/metrics body must be UTF-8");
+
+        // Per-exchange counter — success for a healthy exchange, non-success
+        // for the failing one.
+        assert!(
+            body.contains(r#"xrc_exchange_fetch_total{exchange="Coinbase""#),
+            "missing per-exchange series for Coinbase\n{body}"
```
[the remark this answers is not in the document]
mbjorkqvist: Good point — applied in 418b1b4. The Coinbase assertion now requires the full `{exchange="Coinbase",kind="crypto",outcome="success"}` label set, and the KuCoin failure-line predicate now also requires `kind="crypto"` (in addition to `exchange="KuCoin",` and `outcome != "success"`). A regression that silently drops or renames `kind` or `outcome` will now fail the test instead of slipping through on a name-only prefix match.

src/xrc-tests/src/tests/metrics_endpoint.rs
gregorydemay: Thanks for the explanation and the ticket!

gregorydemay (approved):

src/xrc/src/api/metrics.rs
gregorydemay: Honestly it's not clear to me what Claude tried to achieve even after reading this explanation twice. This is not really a blocker since this part of the code is no longer in the PR.

PR #315, opened 2026-06-01
fix: DEFI-2810: Tune metrics help message character

mbjorkqvist:
The em dash character in the metrics help was showing up as `â€”`. Replace it with a hyphen (`-`).


mbjorkqvist:
See https://uf6dk-hyaaa-aaaaq-qaaaq-cai.raw.icp0.io/metrics (while the current canister version is still deployed).

ielashi (approved):

PR #319, opened 2026-06-01
test: DEFI-2845: pin actual 2-symbol stablecoin selection behaviour

mbjorkqvist:
## Purpose

Adds a unit test that pins the **actual** behaviour of `get_stablecoin_rate` with the production stablecoin set `STABLECOIN_BASES = [USDS, USDC]`, to document a divergence between the implementation and the design slides.

The slides describe a three-input median over `{median_usdc, median_dai, median_usdt = 1}` that rejects a single depegged stablecoin. The code does **not** do this: it takes `median_in_set` over only the two real stablecoin medians, with **no synthetic `USDT = 1` anchor**. With two inputs there is no true middle, so a depegged stablecoin is not rejected — it is selected and flows into the returned USDT/USD rate.

The test uses the slides' own "DAI depegged" example (USDS/USDT median `0.80`, USDC/USDT median `0.99`) and asserts the code returns **USDT/USD = 1/0.80 = 1.25** (`1_250_000_000`), **not** the slides' `1/0.99 = 1.01`. This both confirms the analysis and guards the current behaviour; it relates to the fragility discussion in DEFI-2845.


src/xrc/src/stablecoin.rs
```diff
@@ -451,4 +453,151 @@ mod test {
...
+                    symbol: "USDT".to_string(),
+                    class: AssetClass::Cryptocurrency,
+                },
+                1647734400,
+                &[rate],
+                1,
+                1,
+                None,
+            )
+        };
```
gregorydemay: nit: this is also defined in the previous test. I would either define it as a standalone test helper function, or merge both tests to avoid code duplication

src/xrc/src/stablecoin.rs
```diff
@@ -451,4 +453,151 @@ mod test {
...
+        let low = make("LOW", 950_000_000); // 0.95 (e.g. a depeg)
+        let mid = make("MID", 1_000_000_000); // 1.00
+        let high = make("HIGH", 1_002_000_000); // 1.002
+
+        // The middle value (1.00) is selected regardless of order, so the
+        // depegged 0.95 outlier can never win.
+        let usd = crate::api::usd_asset();
+        let a = get_stablecoin_rate(&[low.clone(), mid.clone(), high.clone()], &usd).expect("a");
+        let b = get_stablecoin_rate(&[high, low, mid], &usd).expect("b");
```
gregorydemay: nit: just unwrap? The string in expect does not seem that informative

```suggestion
        let a = get_stablecoin_rate(&[low.clone(), mid.clone(), high.clone()], &usd).unwrap();
        let b = get_stablecoin_rate(&[high, low, mid], &usd).unwrap();
```

gregorydemay (approved):
Thanks @mbjorkqvist for digging into this!

src/xrc/src/stablecoin.rs
```diff
@@ -451,4 +453,151 @@ mod test {
...
+    /// median_dai, median_usdt = 1}` that rejects a single depegged
+    /// stablecoin. The implementation does NOT match that: it takes
+    /// `median_in_set` over only the two real stablecoin medians and injects
+    /// no synthetic `USDT = 1` anchor. With exactly two inputs there is no true
+    /// "middle", so a depegged stablecoin is not rejected — it is selected and
+    /// flows straight into the returned USDT/USD rate.
+    ///
+    /// Using the slides' "DAI depegged" example (USDS/USDT median 0.80,
+    /// USDC/USDT median 0.99), the code returns USDT/USD = 1 / 0.80 = 1.25,
+    /// NOT the slides' 1 / 0.99 = 1.01.
```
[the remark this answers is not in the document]
mbjorkqvist: Clarified — added a note that USDS is the on-chain symbol that replaced DAI, and dropped the DAI references in favour of the actual USDS symbol used by the test.

src/xrc/src/stablecoin.rs
```diff
@@ -451,4 +453,151 @@ mod test {
...
+    }
+
+    /// Documents that with exactly TWO stablecoin symbols the selection is
+    /// ORDER-DEPENDENT. `median_in_set` has no true middle for two values: they
+    /// are equidistant from their midpoint, so the tie is broken toward the
+    /// FIRST entry (it only replaces on a strictly-smaller distance). Because
+    /// real rates are even-valued in RATE_UNIT terms, that tie is the normal
+    /// case — so the first-listed stablecoin is effectively the default anchor,
+    /// almost regardless of the actual values. Swapping the order changes which
+    /// stablecoin becomes the USDT/USD anchor, even when both are healthy.
```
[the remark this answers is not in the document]
mbjorkqvist: Agreed, that over-claimed. Reworded: the tie occurs when the two medians are equidistant from their integer midpoint, i.e. when their sum is even; the first entry then wins. Removed the rate-parity claim.

src/xrc/src/stablecoin.rs
mbjorkqvist: Done — extracted a module-level `stablecoin_rate(symbol, rate)` helper, now used by all three selection-behaviour tests.

src/xrc/src/stablecoin.rs
mbjorkqvist: Done — switched these to `.unwrap()` as suggested (and the similar ones in the two-symbol test for consistency).

src/xrc/src/stablecoin.rs
```diff
@@ -3,7 +3,9 @@ use ic_xrc_types::{Asset, ExchangeRateError};
 use crate::utils::{median, median_in_set};
 use crate::QueriedExchangeRate;
 
-/// At least 2 stablecoin rates with respect to a third stablecoin are needed to determine if a rate is off.
+/// At least 2 stablecoin rates - each quoted against the quote asset USDT (the
+/// "third stablecoin") - are needed to determine if a rate is off. USDT is the
+/// denominator, not a candidate in the median; see `get_stablecoin_rate`.
```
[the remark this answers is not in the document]
mbjorkqvist: Good point — the function is generic over the quote asset (it only requires all rates to share one). Reworded to 'each quoted against the same quote asset (USDT in production)' and noted the shared quote is the denominator, not a candidate in the median.

PR #320, opened 2026-06-02
fix: DEFI-2845: order STABLECOIN_BASES most-trusted-first ([USDC, USDS])

mbjorkqvist:
## Purpose

A zero-cost interim mitigation for the two-symbol stablecoin selection (DEFI-2845).

With exactly two stablecoin symbols, `median_in_set` has no true middle: the two per-symbol medians are equidistant from their midpoint, so the (almost-always) tie is broken toward the **first** entry. Listing `USDS` first therefore made the thin, depeg-prone USDS the **default USD anchor** — and a USDS depeg would be silently baked into every crypto/USD rate (see the worked 15% scenario in DEFI-2845).

Reordering to `[USDC, USDS]` makes the robust, deep USDC the default tie-winner, so a USDS depeg is avoided on the common path.

## Caveats / scope

- This only **shifts** which coin's depeg propagates (USDC-first means a USDC depeg is always selected); it's an improvement only because USDC is far more reliable and liquid than USDS. It is **not** a fix.
- It is **superseded by the FDUSD PR (#318)**: with an odd set (3 symbols) `median_in_set` returns the true middle by value and order no longer matters, so this change becomes inert.

## Merge order

Merge this **before** #318 if there's any gap before FDUSD ships — that's the only window in which it adds value. The two touch the same line (`STABLECOIN_BASES`), so #318 should rebase on top to `[USDC, USDS, FDUSD]`. If #318 is going out imminently, this PR can simply be closed as superseded.

No behaviour change in the e2e scenarios (their median sums are odd, so the lower value already wins regardless of order); all unit and e2e tests pass unchanged.


src/xrc/src/api.rs
```diff
@@ -27,7 +27,7 @@ use candid::Principal;
 use futures::future::join_all;
 
 /// The expected base rates for stablecoins.
-const STABLECOIN_BASES: &[&str] = &[USDS, USDC];
+const STABLECOIN_BASES: &[&str] = &[USDC, USDS];
```
gregorydemay, line 30: nit: maybe add a unit test: a USDS depeg is ok?

gregorydemay (approved):

src/xrc/src/api.rs
mbjorkqvist: Done in `bc2ee4e` — added `two_symbol_set_tolerates_a_usds_depeg_with_usdc_first` in `stablecoin.rs`. It reuses the same 0.80/0.99 depeg scenario but lists USDC first (production order), asserting the tie breaks toward the healthy USDC → USDT/USD = 1/0.99 = 1.01, so the depegged USDS (which would give 1.25) is not selected. I also refreshed the now-stale `[USDS, USDC]` annotations on the existing `two_symbol_set_does_not_reject_a_depegged_stablecoin` test, reframing it as the first-listed-coin depeg case that this reorder avoids.

PR #324, opened 2026-06-05
fix: DEFI-2860: send a curl-style User-Agent for ReserveBankOfAustralia outcalls

mbjorkqvist:
## Purpose

`IC_XRC_ForexSourceSilent` has been firing for `ReserveBankOfAustralia`: zero successful forex fetches since the last upgrade, while every other forex source is healthy.

The cause is at the HTTP edge, not response size. The rba.gov.au WAF returns HTTP 403 for the canister's default `User-Agent: "Exchange Rate Canister"` (and for browser-like UAs, and for a request with no `User-Agent` header at all), but accepts a plain `curl`-style UA. Every outcall is therefore rejected with 403 → `http_error`, no successes.

This PR overrides the User-Agent for `ReserveBankOfAustralia` only, so the other sources keep using the default UA they already work with. Because `User-Agent` is a singleton header, it gets a first-class per-source override path on the forex trait rather than being routed through the general additional-headers hook.

## Caveat — verify on-subnet

The accepted UA was confirmed only from a single developer machine, not the 13-replica subnet that performs the real outcall, and WAFs can also key on source IP / request fingerprint. The chosen UA must be confirmed to produce successful fetches from the canister (`xrc_forex_fetch_total` success advancing) before closing.

### DIFF
--- a/src/xrc/src/http.rs
+++ b/src/xrc/src/http.rs
@@ -134,37 +134,46 @@ impl CanisterHttpRequest {
 mod test {
     use super::*;
 
+    /// Collects the values of every header whose name matches `name`
+    /// (case-insensitive). Lets tests assert by header name instead of relying
+    /// on the position or total count of headers, and confines the deprecated
+    /// field access to one place.
+    fn header_values(request: &CanisterHttpRequest, name: &str) -> Vec<String> {
+        // TODO(DEFI-2648): Migrate to non-deprecated.
+        #[allow(deprecated)]
+        request
+            .args
+            .headers
+            .iter()
+            .filter(|header| header.name.eq_ignore_ascii_case(name))
+            .map(|header| header.value.clone())
+            .collect()
+    }
+
     /// A new request carries the default `User-Agent` header.
     #[test]
     fn default_user_agent() {
         let request = CanisterHttpRequest::new();
-        // TODO(DEFI-2648): Migrate to non-deprecated.
-        #[allow(deprecated)]
-        let headers = &request.args.headers;
-        assert_eq!(headers.len(), 1);
-        // TODO(DEFI-2648): Migrate to non-deprecated.
-        #[allow(deprecated)]
-        {
-            assert_eq!(headers[0].name, "User-Agent");
-            assert_eq!(headers[0].value, "Exchange Rate Canister");
-        }
+        assert_eq!(
+            header_values(&request, "User-Agent"),
+            vec!["Exchange Rate Canister".to_string()]
+        );
     }
 
-    /// `add_headers` appends headers whose name is not already present.
+    /// `add_headers` appends a header whose name is not already present, leaving
+    /// the default `User-Agent` in place.
     #[test]
     fn add_headers_appends_new_header() {
         let request = CanisterHttpRequest::new()
             .add_headers(vec![("Accept".to_string(), "application/json".to_string())]);
-        // TODO(DEFI-2648): Migrate to non-deprecated.
-        #[allow(deprecated)]
-        let headers = &request.args.headers;
-        assert_eq!(headers.len(), 2);
-        // TODO(DEFI-2648): Migrate to non-deprecated.
-        #[allow(deprecated)]
-        {
-            assert_eq!(headers[1].name, "Accept");
-            assert_eq!(headers[1].value, "application/json");
-        }
+        assert_eq!(
+            header_values(&request, "Accept"),
+            vec!["application/json".to_string()]
+        );
+        assert_eq!(
+            header_values(&request, "User-Agent"),
+            vec!["Exchange Rate Canister".to_string()]
+        );
     }
 
     /// `add_headers` replaces a header with a matching (case-insensitive) name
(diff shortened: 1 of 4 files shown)


src/xrc/src/http.rs
```diff
@@ -122,3 +129,60 @@ impl CanisterHttpRequest {
...
+        // TODO(DEFI-2648): Migrate to non-deprecated.
+        #[allow(deprecated)]
+        let headers = &request.args.headers;
+        assert_eq!(headers.len(), 1);
+        // TODO(DEFI-2648): Migrate to non-deprecated.
+        #[allow(deprecated)]
+        {
+            assert_eq!(headers[0].name, "User-Agent");
+            assert_eq!(headers[0].value, "Exchange Rate Canister");
+        }
```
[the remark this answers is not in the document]
mbjorkqvist: Good point — refactored to assert by header name via a new `header_values(request, name)` test helper, so this no longer depends on header position or total count: it asserts `header_values(&request, "User-Agent") == ["Exchange Rate Canister"]`. Done in 169b934.

src/xrc/src/http.rs
```diff
@@ -122,3 +129,60 @@ impl CanisterHttpRequest {
...
+        // TODO(DEFI-2648): Migrate to non-deprecated.
+        #[allow(deprecated)]
+        let headers = &request.args.headers;
+        assert_eq!(headers.len(), 2);
+        // TODO(DEFI-2648): Migrate to non-deprecated.
+        #[allow(deprecated)]
+        {
+            assert_eq!(headers[1].name, "Accept");
+            assert_eq!(headers[1].value, "application/json");
+        }
```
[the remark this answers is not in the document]
mbjorkqvist: Agreed. The test now looks both headers up by name instead of index/length — asserting the `Accept` header is present with its value and that the default `User-Agent` is still present. 169b934.

src/xrc/src/http.rs
```diff
@@ -122,3 +129,60 @@ impl CanisterHttpRequest {
...
+        // TODO(DEFI-2648): Migrate to non-deprecated.
+        #[allow(deprecated)]
+        let headers = &request.args.headers;
+        assert_eq!(headers.len(), 1);
+        // TODO(DEFI-2648): Migrate to non-deprecated.
+        #[allow(deprecated)]
+        {
+            assert_eq!(headers[0].name, "User-Agent");
+            assert_eq!(headers[0].value, "curl/8.0");
+        }
```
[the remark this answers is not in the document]
mbjorkqvist: Agreed, this was the key one. The test now asserts there is exactly one `User-Agent` header carrying the replaced value (`header_values(&request, "User-Agent") == ["curl/8.0"]`), which verifies the no-duplicate guarantee directly, without asserting the total header count. 169b934.

src/xrc/src/http.rs
```diff
@@ -62,16 +62,23 @@ impl CanisterHttpRequest {
         self
     }
 
-    /// Adds HTTP headers for the request
+    /// Adds HTTP headers for the request, replacing any existing header with
+    /// the same (case-insensitive) name. This lets a source override a default
+    /// header such as the `User-Agent` set in [`CanisterHttpRequest::new`].
     pub fn add_headers(mut self, headers: Vec<(String, String)>) -> Self {
```
gregorydemay, line 70: Maybe it would be cleaner to add a `get_user_agent` method on the `Forex` trait? so that one doesn't have to "overwrite" headers? Overwriting headers is usually problematic because HTTP headers can contain multiple values for the same key, so that the semantic of "add_headers" might not be that clear. WDYT?

mbjorkqvist: Good call — done in 448c702. Added a `get_user_agent()` method to the forex trait (default `DEFAULT_USER_AGENT`, RBA overrides it to `curl/8.0`), routed it through the `Forex` enum, and `call_forex` now passes `forex.get_user_agent()` to a dedicated `CanisterHttpRequest::user_agent()` setter. `add_headers` is back to plain append — no overwrite semantics — and since `User-Agent` is a singleton header (RFC 9110) it now has its own first-class override path instead.

gregorydemay: Since the default is already a trait method, do we still need the default in `CanisterHttpRequest::new`? If not, we could simplify the new method `user_agent` to basically just call `add_headers` and remove a few tests. WDYT

gregorydemay (approved):
Thanks @mbjorkqvist ! I think we could maybe simplify further but otherwise LGTM!

src/xrc/src/http.rs
mbjorkqvist: Good question — short answer, I'd keep the default in `new()`, for two reasons.

1. The crypto-exchange path (`call_exchange`) builds with `CanisterHttpRequest::new()` and never calls `.user_agent()` — `IsExchange` has no `get_user_agent()`, so exchanges rely on the seeded default. (If we add `get_user_agent()` to `IsExchange` for symmetry, both paths would set it explicitly — but I'd still keep the default as a backstop: we now have direct evidence that a *missing* UA is itself a 403 trigger on the RBA WAF, so "never send a UA-less outcall" is worth guaranteeing.)
2. Because `new()` seeds the header, `user_agent()` has to replace in place rather than append — otherwise we'd emit a duplicate `User-Agent`.

On "just call `add_headers`": that's exactly why the UA didn't reuse the existing `get_additional_http_request_headers()` hook (the one Italy uses for `Accept`). `Accept` works as a plain append because it isn't pre-seeded and is legitimately additive; `User-Agent` is seeded by `new()` and is a singleton, so appending duplicates it. Making `add_headers` overwrite-by-name (the earlier version) fixed UA but gave the generic hook a surprising semantic that's wrong for additive headers like `Accept` — hence the dedicated singleton path.

The one way `user_agent` *could* collapse into `add_headers`: drop the `new()` default entirely. Then UA becomes "just another header" added via the same hook as `Accept`, with no dedicated method/setter. The trade-off is the footgun above — any source not setting a UA sends none — so I've left the default in.

I added a pointer to this discussion in [the `canhttp` ticket](https://dfinity.atlassian.net/browse/DEFI-2603), to make the handling of headers across the `crypto` and `forex` exchanges consistent.

PR #327, opened 2026-06-08
fix: DEFI-2867: query Poloniex over a closed-minute window

mbjorkqvist:
## Purpose

Poloniex had the highest crypto fetch error rate among the live-minute exchanges in production (~40–50%). It was queried for the **still-forming current minute**, so replicas making the HTTP outcall at slightly different instants saw different bytes and failed to reach consensus — recorded as a failed outcall. This is the same failure mode #314 fixed for Coinbase; Poloniex never got the analogous fix (its recent change #317 was an unrelated dead USDS-USDT stablecoin pair drop).

This requests a short window of already-closed minutes ending one minute before the requested timestamp. Unlike Coinbase/KuCoin, **Poloniex returns candles oldest-first**, so the extractor now selects the **last** (newest) candle in the window. Adds a multi-candle test pinning that ordering.

Direct API probes confirmed Poloniex BTC_USDT and ICP_USDT are healthy (HTTP 200), so this is a query-construction issue, not a missing-asset/geo/rate-limit one.

Implements DEFI-2867.

### DIFF
--- a/src/xrc/src/exchanges.rs
+++ b/src/xrc/src/exchanges.rs
@@ -478,25 +478,47 @@ type PoloniexResponse = Vec<(
     u64,
 )>;
 
+/// Like Coinbase, Poloniex returns the still-forming current minute for a
+/// single-minute query, so replicas making the HTTP outcall at slightly
+/// different instants see different bytes and fail to reach consensus (recorded
+/// as a failed outcall). To avoid this we request a short window of
+/// already-closed minutes ending one minute before the requested timestamp.
+/// Unlike Coinbase/KuCoin, Poloniex returns candles oldest-first, so
+/// `extract_rate` selects the last (newest) candle in the window. The window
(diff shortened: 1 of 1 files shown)


mbjorkqvist:
Closing: investigation refuted the premise. Empirical probing (parallel bursts + time-series, 2026-06) shows Poloniex DOES forward-fill the current minute (stable open, never empty for BTC and ICP), so the still-forming-minute consensus-failure rationale does not apply — this closed-window change would not fix the production errors. Poloniex's real errors are requests for pairs it does not list (XAUT_USDT empty body, USDC_USDT invalid symbol, etc.), the same class as the Coinbase ICP delisting, which DEFI-2868's listing-gating fixes. Tracked under DEFI-2868; DEFI-2867 closed as superseded. (Contrast PR #326/DEFI-2866 for KuCoin, which does NOT forward-fill and is the correct fix — kept open.)

PR #330, opened 2026-06-09
feat: DEFI-2868: add per-exchange listing parsers for USDT pair discovery

mbjorkqvist:
## DEFI-2868 PR series

Part of the **DEFI-2868** series — *discover tradable pairs per exchange and gate crypto queries by listed pairs*. The PRs land in order, with the behaviour change deliberately last, after the discovered sets have been observed populating correctly in production:

1. **#330 — listing parsers** (pure, no wiring) ← *you are here*
2. **#331 — canbench benchmarks + regression gate** (stacked on #330)
3. **#332 — store + daily refresh + metrics** (populate the sets, still no gating)
4. **#337 — gate crypto queries on the discovered sets** (the behaviour change; draft — merge/deploy only after #330–#332 are in prod and observed healthy)

---

## Purpose

First PR in the DEFI-2868 series — *discover tradable pairs per exchange and gate crypto queries by listed pairs*. The crypto path currently queries every exchange for base/USDT even when the pair isn't listed, so unlisted/delisted pairs fail on every request. This PR adds the parsing foundation: reading each exchange's public spot-listing endpoint into the set of bases it currently lists against USDT.

It is deliberately **pure** — no outcalls, storage, or behaviour change, and nothing calls the new code yet.

## What it adds

- Listing-endpoint parsing for all nine exchanges (Coinbase, KuCoin, OKX, GateIo, MEXC, Poloniex, CryptoCom, Bitget, Digifinex), each with its listing URL and tradable rule.
- A result carrying the USDT-tradable base assets plus a total spot-market count — the latter is a structural-health signal for the refresh acceptance guard in a later PR (it stays stable across delistings/migrations, so a sudden drop means a parser break rather than a legitimate change).
- Per-exchange fixtures and unit tests.

## Validation

Parsers were validated against live captures of all nine endpoints; the discovered counts matched expectations — e.g. Coinbase shows ~23 USDT bases with ICP-USDT correctly absent (the delisting this series ultimately fixes), and CryptoCom excludes its ~300 derivatives from the spot-market count.

## Follow-ups (subsequent PRs)

- **#331** — canbench benchmarks for the parsers + a regression gate.
- **#332** — wire a daily refresh timer + stable-memory store + metrics (populate the sets, still no gating).
- **#337** — gate the crypto query path on the discovered sets and remove the hardcoded Coinbase ICP skip.

### DIFF
--- a/src/xrc/src/exchanges.rs
+++ b/src/xrc/src/exchanges.rs
@@ -219,6 +219,10 @@ impl ListedMarket {
 ///   guard: it stays roughly stable across refreshes (even when a venue
 ///   migrates USDT→USD, collapsing `bases`), so a sudden drop indicates a
 ///   parser break or a garbage response rather than a legitimate delisting.
+///   Caveat: MEXC's `defaultSymbols` endpoint enumerates only *tradable*
+///   symbols (see the MEXC impl), so there `total_markets` is weaker as a
+///   stability signal — a mass trading suspension would dent it even though the
+///   response is structurally sound.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct ListedPairs {
     pub bases: BTreeSet<String>,
@@ -648,7 +652,10 @@ type MexcResponse = Vec<(u64, String, String, String, String, String, u64, Strin
 /// MEXC's `defaultSymbols` lists only tradable symbols as concatenated strings
 /// (e.g. `"BTCUSDT"`) with no separator, so the quote can only be recovered for
 /// the suffixes we care about. Non-USDT symbols are kept (so they count toward
-/// `total_markets`) but cannot be split, so their quote is left unknown.
+/// `total_markets`) but cannot be split, so their quote is left unknown. Note
+/// that because this endpoint omits non-tradable symbols, `total_markets` here
+/// tracks the tradable universe rather than the full listing — a weaker
+/// structural-health signal than for other exchanges (see [ListedPairs]).
 #[derive(Deserialize)]
 struct MexcDefaultSymbols {
     data: Vec<String>,
--- a/src/xrc/src/exchanges.rs
+++ b/src/xrc/src/exchanges.rs
@@ -1,3 +1,5 @@
+use std::collections::BTreeSet;
+
 use candid::{decode_args, encode_args, Deserialize, Error as CandidError};
 
 use ic_xrc_types::Asset;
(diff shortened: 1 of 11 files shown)


src/xrc/src/exchanges.rs
```diff
@@ -170,6 +189,79 @@ fn extract_rate<R: DeserializeOwned>(
...
+///   guard: it stays roughly stable across refreshes (even when a venue
+///   migrates USDT→USD, collapsing `bases`), so a sudden drop indicates a
+///   parser break or a garbage response rather than a legitimate delisting.
+///   Caveat: MEXC's `defaultSymbols` endpoint enumerates only *tradable*
+///   symbols (see the MEXC impl), so there `total_markets` is weaker as a
+///   stability signal — a mass trading suspension would dent it even though the
+///   response is structurally sound.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ListedPairs {
+    pub bases: BTreeSet<String>,
```
gregorydemay, line 228: nit: maybe we want a proper `TokenSymbol(String)` struct so that `TokenSymbol` is responsible for the normalization (uppercase, ascii, trim white space, etc)

src/xrc/src/exchanges.rs
```diff
@@ -871,6 +1232,110 @@ mod test {
         assert!(matches!(extracted_rate, Ok(rate) if rate == 11_357_000_000));
     }
 
+    /// Asserts the common shape of every listing fixture: each holds four spot
+    /// markets — two tradable USDT pairs (BTC, ETH), one tradable non-USDT pair,
+    /// and one non-tradable USDT pair — so a correct parser yields exactly
+    /// `{BTC, ETH}` as the gating set while still counting all four (or, for
+    /// Crypto.com, four spot pairs plus a derivative that must be excluded) in
+    /// `total_markets`.
+    fn assert_lists_btc_and_eth(exchange: &impl IsExchange, file: &str) {
```
gregorydemay, line 1277: nit: since the list of supported pairs does not seem that big in the test-data.json maybe it would be better to assert on the exact listed assets for each exchange instead of a strict subset?

gregorydemay (approved):
Thanks @mbjorkqvist !

src/xrc/src/exchanges.rs
mbjorkqvist: Good idea! Asset symbols are plain `String`/`&str` everywhere else in this module - `Asset.symbol`, the `USDT` const, `format_asset`'s `.to_uppercase()` - so a `TokenSymbol` newtype only in the listing path would be an inconsistent island, and threading it crate-wide is a larger refactor than fits this staged stack. The normalization gating depends on is already centralized in one spot (`extract_listed_pairs`: `.to_uppercase()` on the base, `eq_ignore_ascii_case` on the quote). I filed [DEFI-2884](https://dfinity.atlassian.net/browse/DEFI-2884) to introduce `TokenSymbol` crate-wide as a follow-up.

src/xrc/src/exchanges.rs
mbjorkqvist: Maybe I'm not understanding your question, but I don't see what we should be changing here? In the test-data json files, all exchanges return only two supported pairs: `BTC` and `ETH` (as base assets), with `USDT` as the quote asset, and we assert on those exactly. The other assets are not supported (e.g., `LUMIA/BTC` for bitget), so the `IsExchange` parser filters them out. We have `total_markets` so that we can alert if there's a significant change, and look into it (manually) to see if we need to somehow adapt (e.g., remove the exchange, update the parser, add/change a symbol).

gregorydemay: > The other assets are not supported (e.g., LUMIA/BTC for bitget), so the IsExchange parser filters them out.

Ha I see, I missed that part. Thanks for the explanation!

PR #331, opened 2026-06-09
test: DEFI-2868: canbench benchmarks for listing parsers and regression gate

mbjorkqvist:
## Purpose

Stacked on #330. De-risks the DEFI-2868 listing-refresh design by **measuring the instruction cost of the per-exchange listing parse**, which will run inside the HTTP-outcall transform — a single, non-DTS execution on the feeless system subnet, so its instruction count (not cycles) is the binding constraint when refreshing the up-to-~1.2 MiB listing endpoints.

> Review/merge #330 first; this targets that branch and will be retargeted to `main` once #330 lands.

## What it adds

- **canbench harness** (`src/xrc/src/benchmarks.rs`, gated behind the optional `canbench-rs` feature) that benchmarks the production `extract_listed_usdt_bases` on real, full-size captures of all nine endpoints. Worst case is **OKX at ~85M instructions** (~1–4% of a single-slice budget), so the parse fits with large margin — no compression or pagination needed.
- **CI parity with `dfinity/dex`**: a required `benchmark` job that fails whenever the committed `canbench_results.yml` is out of date (any regression/improvement/new benchmark), forcing perf changes to be re-baselined and reviewed in the diff; plus a separate `canbench-post-comment` workflow that posts the instruction delta vs the base branch as a PR comment.

canbench is pinned to `0.4.1` and the build is `--locked` so CI reproduces the committed counts exactly. The benchmarks call the production parser directly (no duplicated code), so a regression in the canister's listing parse is what the check catches. Normal canister builds and tests are unaffected — the feature is off by default.

## Notes

- The benchmark job runs on push and PR (not PR-only) so it is never *skipped*, keeping xrc's skip-intolerant `checks-pass` gate green without touching branch protection.
- The ~5 MiB of committed captures could be trimmed to the binding cases (OKX + GateIo) before merge if preferred.


src/xrc/benches/data/bitget.json
```diff
@@ -0,0 +1 @@
...mount":"0","maxTradeAmount":"900000000000000000000","takerFeeRate":"0.001","makerFeeRate":"0.001","pricePrecision":"3","quantityPrecision":"2","quotePrecision":"5","status":"online","minTradeUSDT":"1","buyLimitPriceRatio":"0.02","sellLimitPriceRatio":"0.02","areaSymbol":"no","orderQuantity":"200","openTime":"1692100800000","offTime":"","maxLimitOrderValue":"20000000","maxMarketOrderValue":"1000000"},{"symbol":"RSLVUSDT","baseCoin":"rSLV","quoteCoin":"USDT","minTradeAmount":"0","maxTradeAmount":"900000000000000000000","takerFeeRate":"0.001","makerFeeRate":"0.001","pricePrecision":"2","quantityPrecision":"2","quotePrecision":"4","status":"online","minTradeUSDT":"10","buyLimitPriceRatio":"0.1","sellLimitPriceRatio":"0.1","areaSymbol":"yes","orderQuantity":"200","openTime":"1780656001755","offTime":"","maxLimitOrderValue":"100000","maxMarketOrderValue":"100000"},{"symbol":"RUNGUSDT","baseCoin":"rUNG","quoteCoin":"USDT","minTradeAmount":"0","maxTradeAmount":"900000000000000000000","takerFeeRate":"0.001","makerFeeRate":"0.001","pricePrecision":"2","quantityPrecision":"2","quotePrecision":"4","status":"online","minTradeUSDT":"20","buyLimitPriceRatio":"0.1","sellLimitPriceRatio":"0.1","are
```
gregorydemay, line 1: Here for a lack of a better place. Maybe add a README in this folder explaining how those data were obtained (maybe a little bash code snippet would help)

.github/workflows/canbench-post-comment.yml
```diff
@@ -0,0 +1,47 @@
...
+jobs:
+  download-results:
+    if: github.event.workflow_run.event == 'pull_request'
+    runs-on: ubuntu-22.04
+    outputs:
+      matrix: ${{ steps.set-benchmarks.outputs.matrix }}
+      pr_number: ${{ steps.set-benchmarks.outputs.pr_number }}
+    steps:
+      - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
+      - uses: dawidd6/action-download-artifact@09f2f74827fd3a8607589e5ad7f9398816f540fe
```
gregorydemay: This was not that stable in the past, can be replaced by the standard Github action, see https://github.com/dfinity/dex/pull/132

gregorydemay (approved):
Thanks @mbjorkqvist !

scripts/ci_run_benchmark.sh
```diff
@@ -0,0 +1,102 @@
...
+
+CANBENCH_OUTPUT=/tmp/canbench_output_${CANBENCH_JOB_NAME}.txt
+
+CANBENCH_RESULTS_FILE="$CANISTER_PATH/canbench_results.yml"
+BASELINE_BRANCH_RESULTS_FILE="$BASELINE_BRANCH_DIR/$CANBENCH_RESULTS_FILE"
+
+CANBENCH_RESULTS_CSV_FILE="/tmp/canbench_results_${CANBENCH_JOB_NAME}.csv"
+
+# Install canbench (skip if a matching version is already cached).
+command -v canbench >/dev/null 2>&1 || cargo install --version 0.4.1 --locked canbench
```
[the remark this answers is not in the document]
mbjorkqvist: Good catch — the install step now verifies the exact version instead of just the binary's presence: `canbench --version | grep -qw 0.4.1`, reinstalling with `--force` on mismatch. A cached or preinstalled build of a different version can no longer silently change the output format or measured instruction counts. Fixed in 72fb397.

src/xrc/Cargo.toml
```diff
@@ -24,6 +24,11 @@ serde_bytes = "0.11.19"
 serde-xml-rs = "0.6.0"
 strum = { workspace = true }
 
+# Only pulled in for the `canbench` instruction benchmarks of the per-exchange
+# listing parsers (see benches/ and main.rs). The implicit `canbench-rs` feature
+# created by this optional dependency gates them out of normal canister builds.
+canbench-rs = { version = "0.4", optional = true }
```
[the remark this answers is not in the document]
mbjorkqvist: Agreed — pinned to `=0.4.1` (with a comment on the rationale). The lockfile already resolved to 0.4.1, so this just makes the intended pin explicit and stops an accidental 0.4.x bump from silently shifting the benchmarked instruction counts. Fixed in 72fb397.

src/xrc/src/benchmarks.rs
```diff
@@ -0,0 +1,48 @@
+//! Instruction benchmarks for the per-exchange listing parsers, measured with
+//! `canbench` (see `canbench.yml`).
+//!
+//! The listing parse runs inside the HTTP-outcall transform — a single,
+//! non-DTS execution whose instruction budget is the binding constraint when
+//! refreshing the (up to ~1.2 MiB) listing endpoints — so these benchmark
+//! `extract_listed_usdt_bases` on real, full-size captures (`benches/data/`,
+//! the worst case being OKX at ~1.2 MiB). They call the production parser
```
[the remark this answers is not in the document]
mbjorkqvist: Fixed — reworded to "so this module benchmarks `extract_listed_usdt_bases` …". 72fb397.

src/xrc/benches/data/bitget.json
mbjorkqvist: Added `benches/data/README.md`: a table mapping each capture to its exchange listing endpoint (the `listing_url()` the refresh fetches) plus a curl snippet to regenerate them. It also notes keeping the responses byte-for-byte as served (no reformatting/minifying), since reformatting would change the parse cost the benchmark measures. 72fb397.

.github/workflows/canbench-post-comment.yml
mbjorkqvist: Done — swapped `dawidd6/action-download-artifact` for `actions/download-artifact@3e5f45b… # v8.0.1` (`run-id` + `github-token`), mirroring dfinity/dex#132. The per-artifact-name subdirectory layout that `ci_download_canbench_artifacts.sh` globs is preserved. 72fb397.

PR #332, opened 2026-06-10
feat: DEFI-2868: discover listings periodically - store, refresh, metrics

mbjorkqvist:
## DEFI-2868 PR series

Part of the **DEFI-2868** series — *discover tradable pairs per exchange and gate crypto queries by listed pairs*. The PRs land in order, with the behaviour change deliberately last, after the discovered sets have been observed populating correctly in production:

1. **#330 — listing parsers** (pure, no wiring)
2. **#331 — canbench benchmarks + regression gate** (stacked on #330)
3. **#332 — store + daily refresh + metrics** (populate the sets, still no gating) ← *you are here*
4. **#337 — gate crypto queries on the discovered sets** (the behaviour change; draft — merge/deploy only after #330–#332 are in prod and observed healthy)

---

## Purpose

Stacked on #330. Second step of DEFI-2868: populate and expose the per-exchange USDT listings discovered by #330's parsers, **without changing query behaviour yet**. Gating the crypto path on the discovered sets is the next PR, **#337 (PR3)** — this one lets the sets be watched in prod first.

> Review/merge #330 first; this targets that branch and will be retargeted to `main` once #330 lands.

## What it adds

- **`ListingStore`** (`exchange → { bases, total_markets, last_success_secs }`) with a **structural acceptance guard**: a fetched listing replaces the stored one only if it has at least a floor of total markets and at least half the previously accepted total; otherwise the last-known-good listing is kept. The guard is judged on *total* markets, never the USDT subset, so a USDT→USD migration is accepted while a parser break is rejected.
- **Listing outcall + consensus transform** — the transform parses the (up to ~1.2 MiB) body and replaces it with a small canonical `(sorted bases, total)` payload, so replicas reach consensus on that, not the raw listing.
- **Daily refresh task** on the heartbeat (mirrors the forex update): one outcall per available exchange, each result run through the guard into the store.
- **Four metrics** — `xrc_exchange_listed_usdt_pairs`, `xrc_exchange_listing_total_markets`, `xrc_exchange_listing_last_success_seconds` (gauges) and `xrc_exchange_listing_rejected_total` (counter) — backing the A1/A2/A3 alerts.
- **Stable-memory persistence** — `pre_upgrade`/`post_upgrade` save/restore the store; the listing store is decoded as a trailing `Option` so the first upgrade from the old single-`ForexRateStore` layout doesn't trap.

## Not in this PR

- **No gating.** `get_cryptocurrency_usdt_rate` is unchanged; the store is populated and observable only. The gating read (with fail-open on a missing/stale listing) is **#337**. Once this PR is in production, the four metrics above are how reviewers confirm the discovered sets are correct and refreshing before gating is enabled in #337.

## Verification

clippy `-D warnings` clean; 247 lib unit tests (incl. the guard's worked scenarios, the refresh task with a mock source, the metrics, and the upgrade-migration decode); candid-compat green; production Wasm builds.

### DIFF
--- a/src/xrc/src/api/metrics.rs
+++ b/src/xrc/src/api/metrics.rs
@@ -188,7 +188,7 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
     encode_labeled_counter_family(
         w,
         MetricName::ExchangeListingRejectedTotal,
-        "Total per-exchange listing refreshes not applied, labeled by reason: 'fetch' (HTTP outcall failed - exchange unreachable) or 'guard' (a 200 that failed the structural acceptance guard - API change or parser break). A rising 'guard' rate points at a parser/API issue; a rising 'fetch' rate at connectivity.",
+        "Total per-exchange listing refreshes not applied, labeled by reason: 'fetch' (no listing could be fetched and parsed - an HTTP outcall error, transform trap, or candid encode/decode failure) or 'guard' (a 200 that failed the structural acceptance guard - API change or parser break). A rising 'guard' rate points at a parser/API issue; a rising 'fetch' rate at connectivity or a malformed/oversized response.",
     )?;
 
     Ok(())
--- a/src/xrc/src/api/metrics.rs
+++ b/src/xrc/src/api/metrics.rs
@@ -188,7 +188,7 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
     encode_labeled_counter_family(
         w,
         MetricName::ExchangeListingRejectedTotal,
-        "Total per-exchange listing refreshes rejected by the structural acceptance guard or that failed to fetch; a rising rate indicates an exchange API change or a parser break.",
+        "Total per-exchange listing refreshes not applied, labeled by reason: 'fetch' (HTTP outcall failed - exchange unreachable) or 'guard' (a 200 that failed the structural acceptance guard - API change or parser break). A rising 'guard' rate points at a parser/API issue; a rising 'fetch' rate at connectivity.",
     )?;
 
     Ok(())
--- a/src/xrc/src/api/metrics.rs
+++ b/src/xrc/src/api/metrics.rs
@@ -170,6 +170,26 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
         MetricName::StablecoinSymbolRatesReceived,
         "Count of individual exchange rate samples received per stablecoin symbol; its rate drops to zero when an upstream symbol stops returning data (e.g. after a rebrand).",
     )?;
+    encode_labeled_gauge_family(
+        w,
+        MetricName::ExchangeListedUsdtPairs,
+        "Number of base assets each exchange currently lists against USDT, from the last accepted listing refresh (the set the crypto path is gated on).",
+    )?;
+    encode_labeled_gauge_family(
+        w,
+        MetricName::ExchangeListingTotalMarkets,
+        "Total spot markets parsed (across all quotes) per exchange in the last accepted listing refresh; a structural-health signal that stays stable across delistings and USDT-to-USD migrations.",
+    )?;
+    encode_labeled_gauge_family(
+        w,
+        MetricName::ExchangeListingLastSuccessSeconds,
+        "Unix timestamp (seconds) of the most recent accepted listing refresh per exchange.",
+    )?;
+    encode_labeled_counter_family(
+        w,
+        MetricName::ExchangeListingRejectedTotal,
+        "Total per-exchange listing refreshes rejected by the structural acceptance guard or that failed to fetch; a rising rate indicates an exchange API change or a parser break.",
+    )?;
 
     Ok(())
 }
--- a/src/xrc/src/listings.rs
+++ b/src/xrc/src/listings.rs
@@ -84,7 +84,10 @@ impl ListingStore {
             // Guard on TOTAL markets, not the USDT subset: a venue migrating
             // USDT->USD keeps total roughly stable (accepted) while its USDT
             // bases collapse, whereas a parser break collapses total (rejected).
-            let min_total = (previous.total_markets as f64 * MIN_RETAINED_FRACTION) as u64;
+            // Round the floor up: truncating would let an odd previous total
+            // accept a refresh just under the retained fraction (e.g. previous
+            // 101 -> floor 50, and 50/101 < 0.5 would slip through).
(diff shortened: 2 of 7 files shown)


src/xrc/src/periodic.rs
```diff
@@ -283,6 +292,164 @@ fn get_next_run_timestamp(timestamp: u64) -> u64 {
     (timestamp - (timestamp % SIX_HOURS)) + SIX_HOURS
 }
 
+// The per-exchange listings are refreshed roughly once a day.
+const LISTING_REFRESH_INTERVAL: u64 = ONE_DAY_SECONDS;
+
+// When a refresh accepts no exchange at all (every fetch failed or was
+// rejected), retry after this much shorter interval instead of waiting a full
+// day, so a transient outage doesn't cost a day of staleness.
+const LISTING_RETRY_INTERVAL: u64 = ONE_HOUR_SECONDS;
```
gregorydemay: Did you consider the option of having something per exchange instead? Each exchange gets its own listing task, programmed independently of the other. In particular if one exchange has an outage (but the others are fine), then this exchange could be retried more aggressively (like every hour). The case where all exchanges have an outage at the same time seems unlikely to me. WDYT?

src/xrc/src/periodic.rs
```diff
@@ -283,6 +292,164 @@ fn get_next_run_timestamp(timestamp: u64) -> u64 {
...
+
+/// Refreshes the per-exchange listing store once the daily interval has
+/// elapsed. Each exchange is independent: an accepted listing replaces the
+/// stored one, while a rejected (guard) or failed (HTTP) fetch leaves the
+/// last-known-good listing in place.
+async fn update_listing_store(
+    timestamp: u64,
+    listing_sources: &impl ListingSources,
+) -> UpdateListingStoreResult {
+    let next_run_at = get_next_listing_run_scheduled_at_timestamp();
```
gregorydemay: I'm wondering if we should not right here set already the next timer. Otherwise, if there is a panic in the logic below (after the await), we will keep retrying on every heartbeat. WDYT?

src/xrc/src/periodic.rs
```diff
@@ -619,6 +786,255 @@ mod test {
...
+        fn accepted_listing_populates_store() {
+            ready();
+            let sources = MockListingSources {
+                results: vec![(
+                    "ListingTestAccept".to_string(),
+                    MockResult::Listed(listed(&["BTC", "ICP"], 300)),
+                )],
+            };
+            let result = update_listing_store(1_000, &sources)
+                .now_or_never()
```
gregorydemay, line 933: nice, I didn't know this trick.

gregorydemay (approved):
Thanks @mbjorkqvist , come minor comments but otherwise LGTM!

src/xrc/src/listings.rs
```diff
@@ -0,0 +1,246 @@
...
+            // USDT->USD keeps total roughly stable (accepted) while its USDT
+            // bases collapse, whereas a parser break collapses total (rejected).
+            let min_total = (previous.total_markets as f64 * MIN_RETAINED_FRACTION) as u64;
+            if total < min_total {
+                return AcceptOutcome::RejectedTotalDrop {
+                    total,
+                    previous: previous.total_markets,
+                };
+            }
+        }
```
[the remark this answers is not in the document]
mbjorkqvist: Good catch — fixed in 3126fb4. The floor is now rounded up with `.ceil()`, so an odd previous total no longer accepts a refresh just below the fraction (101 → floor 51, rejecting 50/101). Added an odd-total regression test.

src/xrc/src/periodic.rs
mbjorkqvist: Done in b6d51d2. Switched to a per-exchange next-attempt schedule — each exchange is scheduled independently: an accepted refresh rejoins the aligned daily boundary, a failed/rejected one retries after an hour, so a single exchange's outage is retried hourly without pulling the healthy exchanges off their daily cadence, and it snaps back to daily on recovery. Each exchange keeps a single overwritten next-attempt timestamp (never a second queued slot), so a long outage stays at one outcall per interval rather than piling up into multiple simultaneous calls.

src/xrc/src/periodic.rs
mbjorkqvist: Good point — the per-exchange rewrite in b6d51d2 now sets each due exchange's next-attempt to `now + retry` *before* the outcalls, so a trap after the await can't cause a re-fire on every heartbeat. (The thread above has the fuller description.)

src/xrc/src/periodic.rs
```diff
@@ -283,6 +302,204 @@ fn get_next_run_timestamp(timestamp: u64) -> u64 {
...
+            }
+            Err(error) => {
+                // Keep the last-known-good listing and the retry interval
+                // scheduled above on a failed fetch.
+                increment_labeled_counter(
+                    MetricName::ExchangeListingRejectedTotal,
+                    &[(LabelKey::Exchange, &exchange), (LabelKey::Reason, "fetch")],
+                );
+                ic_cdk::println!("{} [listing] {} {}", LOG_PREFIX, exchange, error);
+            }
```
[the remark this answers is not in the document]
mbjorkqvist: Good catch — fixed in 69e6766. The failure log now uses the structured error (`{:?}`/Debug) with a listing-specific prefix (`[listing] <exchange> refresh fetch failed: ...`), so it no longer borrows the rate-fetch wording from `Display`.

src/xrc/src/api/metrics.rs
```diff
@@ -170,6 +170,26 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
...
+    encode_labeled_gauge_family(
+        w,
+        MetricName::ExchangeListingLastSuccessSeconds,
+        "Unix timestamp (seconds) of the most recent accepted listing refresh per exchange.",
+    )?;
+    encode_labeled_counter_family(
+        w,
+        MetricName::ExchangeListingRejectedTotal,
+        "Total per-exchange listing refreshes not applied, labeled by reason: 'fetch' (HTTP outcall failed - exchange unreachable) or 'guard' (a 200 that failed the structural acceptance guard - API change or parser break). A rising 'guard' rate points at a parser/API issue; a rising 'fetch' rate at connectivity.",
+    )?;
```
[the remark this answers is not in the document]
mbjorkqvist: You're right — fixed in 69e6766 by widening the description (the counter and `fetch` label are kept). The help text now reads: 'fetch' (no listing could be fetched and parsed - an HTTP outcall error, transform trap, or candid encode/decode failure). I kept a single `fetch` label rather than splitting into http/candid: on the feeless production subnet the realistic failure is an outcall/transform problem, the variants share the same operator response (it's a connectivity/response-shape issue, not a guard/parser-logic one), and a single label keeps the alert query simple. Happy to split if you'd prefer the finer granularity.

PR #334, opened 2026-06-10
fix: DEFI-2648: migrate monitor-canister off removed ic-cdk call APIs

mbjorkqvist:
Second of the [DEFI-2648](https://dfinity.atlassian.net/browse/DEFI-2648) PRs removing the `#[allow(deprecated)]` shields, after #333 (chrono).

ic-cdk 0.20 **removes** the `api::call` module outright, so the monitor-canister's `call_with_payment` and its `RejectionCode` are migrated now, on 0.19 — ahead of the [DEFI-2878](https://dfinity.atlassian.net/browse/DEFI-2878) bump — so they don't block it.

- `call_with_payment` → the `Call` builder. `unbounded_wait` preserves the old guaranteed-response semantics; the new failure types are mapped into the existing `CallError`.
- A local `RejectionCode` enum is vendored with the same variants, so the hand-maintained `monitor-canister.did` and the candid-encoded stable-log entries stay byte-compatible now that the CDK no longer exposes its own reject-code type.
- The beat's `spawn_017_compat` → `spawn`; `call_xrc` only awaits the XRC inter-canister call, so the task stays within the method lifetime.

[DEFI-2648]: https://dfinity.atlassian.net/browse/DEFI-2648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DEFI-2878]: https://dfinity.atlassian.net/browse/DEFI-2878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/monitor-canister/src/types.rs
```diff
@@ -41,18 +42,77 @@ pub struct Entry {
...
+            // enqueued. Surface it as a transient system error.
+            CallFailed::CallPerformFailed(error) => CallError {
+                rejection_code: RejectionCode::SysTransient,
+                err: error.to_string(),
+            },
+            // We lacked the cycles to perform the call ourselves.
+            CallFailed::InsufficientLiquidCycleBalance(error) => CallError {
+                rejection_code: RejectionCode::CanisterError,
+                err: error.to_string(),
+            },
```
gregorydemay: Can this error happen at all given that this is deployed on a system subnet? I'm ok with leaving the error mapping as is (to be able for example to install that canister on an application subnet) but maybe a comment would be useful here.

gregorydemay (approved):
Thanks @mbjorkqvist !

src/monitor-canister/src/types.rs
mbjorkqvist: Good point, on a system subnet `InsufficientLiquidCycleBalance` should never occur. I've kept the mapping so the canister stays installable on an application subnet, but added a comment as suggested.

PR #335, opened 2026-06-10
fix: release the XRC-call guard when entry encoding fails

mbjorkqvist:
Found while reviewing #334: `call_xrc` returns mid-batch when an `Entry` fails to candid-encode, without clearing `is_calling_xrc`. Every later heartbeat then bails out at the guard, so a single encode failure would silently stop the monitor's sampling until the next upgrade.

Break out of the batch instead, so the epilogue releases the guard and schedules the next run as usual. Pre-existing bug, independent of the #334 migration (the fix applies on `main`).

### DIFF
--- a/src/monitor-canister/src/periodic.rs
+++ b/src/monitor-canister/src/periodic.rs
@@ -179,7 +179,10 @@ async fn call_xrc(xrc_impl: impl Xrc, now_secs: u64) {
             Ok(bytes) => bytes,
             Err(_) => {
                 ic_cdk::println!("Failed to encode Entry");
-                return;
+                // Abandon the batch, but fall through to the epilogue below:
+                // returning here would leave `is_calling_xrc` set forever,
+                // silently halting all future sampling.
+                break;
             }
         };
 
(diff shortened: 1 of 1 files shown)


src/monitor-canister/src/periodic.rs
```diff
@@ -179,7 +179,10 @@ async fn call_xrc(xrc_impl: impl Xrc, now_secs: u64) {
             Ok(bytes) => bytes,
             Err(_) => {
                 ic_cdk::println!("Failed to encode Entry");
-                return;
+                // Abandon the batch, but fall through to the epilogue below:
+                // returning here would leave `is_calling_xrc` set forever,
+                // silently halting all future sampling.
+                break;
```
gregorydemay, line 197: Could we have a test for this behaviour?

gregorydemay (approved):

src/monitor-canister/src/periodic.rs
mbjorkqvist: Added `call_xrc_releases_guard_when_encoding_fails` in c6c47e5. To make the encode-failure path testable, `call_xrc` now delegates to `call_xrc_with_encoder`, which takes an injected encoder (production passes `encode_one`). The test forces every entry to fail encoding and asserts the batch is abandoned without persisting anything, the `is_calling_xrc` guard is released, the next run is scheduled, and a subsequent heartbeat samples again. Reverting `break` → `return` makes the test fail.

PR #336, opened 2026-06-10
test: DEFI-2600 migrate e2e tests to PocketIC (basic_exchange_rates spike)

mbjorkqvist:
Begins migrating the XRC end-to-end tests off the Docker + dfx + nginx harness and onto PocketIC, now that PocketIC can mock HTTPS outcalls — the capability whose absence forced the old setup.

This is the first, foundational step: it ports a single scenario (`basic_exchange_rates`) end-to-end and wires it into CI, proving the approach before the remaining scenarios follow one per PR.

## What this does
- Runs the XRC and a small cycles-forwarding proxy inside PocketIC, answering the canister's outcalls from the existing mock dataset (PocketIC applies the XRC transforms, as the replica does on mainnet). No Docker, nginx, or dfx; the scenario runs in seconds.
- Keeps every request and expected median/standard-deviation assertion unchanged, so behaviour is provably identical to the old harness.
- Adds a CI job (installing the PocketIC server via the `dfinity/pocketic` action) that runs the migrated test and is part of the required checks gate, so it cannot bit-rot. It runs **alongside** the existing dfx e2e job.

## Why a proxy canister
`get_exchange_rate` requires attached cycles, which ingress messages cannot carry. Calls are routed through an in-repo proxy that forwards them with cycles attached — the same role the cycles wallet plays for `dfx canister call --wallet --with-cycles`.

Existing helper crates cover this pattern (`ic-canister-runtime`'s `CyclesWalletRuntime` for the cycles forwarding, and `ic-pocket-canister-runtime` for the outcall mocking), as used by the EVM-RPC canister. They are deliberately **not** adopted here: they require ic-cdk 0.20 and pin pocket-ic to 13.x, which would couple this test migration to the planned ic-cdk bump (DEFI-2878) and forgo the latest PocketIC. Keeping a small in-repo proxy lets the XRC stay on its current ic-cdk and this work stay independent. Once the ic-cdk bump lands, adopting those crates is a reasonable consolidation to revisit.

## Scope
- The other five scenarios, the `caching` redesign, and removal of the Docker/dfx machinery are out of scope here. They follow one per PR, each keeping CI green; the dfx e2e job and Docker harness are retired once all scenarios are ported.
- Unblocks DEFI-2602 (timer instead of heartbeats), which was waiting on a modern replica.


src/xrc-test-proxy/src/main.rs
```diff
@@ -0,0 +1,40 @@
...
+
+#[ic_cdk::init]
+fn init(xrc: Principal) {
+    XRC.with(|cell| cell.replace(Some(xrc)));
+}
+
+#[ic_cdk::update]
+async fn get_exchange_rate(request: GetExchangeRateRequest) -> GetExchangeRateResult {
+    let xrc = XRC.with(|cell| cell.borrow().expect("XRC canister id has not been set"));
+    let response = Call::unbounded_wait(xrc, "get_exchange_rate")
```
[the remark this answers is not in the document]
mbjorkqvist: This compiles as-is — `candid::Principal` is `Copy`, so `cell.borrow().expect(..)` copies the principal out of the `Ref` rather than moving it; there is no borrow/move error. The proxy builds and the PocketIC test runs green in CI, so I have left it unchanged.

src/xrc-tests/src/pocket.rs
```diff
@@ -0,0 +1,182 @@
...
+        pic.install_canister(proxy, proxy_wasm, Encode!(&xrc).unwrap(), None);
+
+        let mut mocks: HashMap<String, ExchangeResponse> = HashMap::with_capacity(responses.len());
+        for response in responses {
+            let url = response.url.clone();
+            let existing = mocks.insert(url.clone(), response);
+            // The URL is the match key; two responses for the same URL would silently shadow,
+            // so flag it as a dataset mistake rather than testing only the last one.
+            debug_assert!(existing.is_none(), "duplicate mock response for URL: {url}");
+        }
```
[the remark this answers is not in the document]
mbjorkqvist: Good point — switched to `assert!` in 7fea0cd so the duplicate-URL guard holds even under `--release` test runs.

src/xrc-tests/src/pocket.rs
```diff
@@ -0,0 +1,182 @@
...
+/// Build a PocketIC reply from a canned [`ExchangeResponse`]. The raw body is returned unchanged;
+/// PocketIC runs the XRC's transform on it (header stripping, rate extraction) as on mainnet.
+fn reply_for(mock: &ExchangeResponse) -> CanisterHttpResponse {
+    // The harness delivers mock responses immediately; it has no notion of response delay yet.
+    // A scenario that sets delay_secs (e.g. caching/misbehavior) would otherwise silently test
+    // nothing, so make the gap announce itself until delay support is added.
+    debug_assert_eq!(
+        mock.delay_secs, 0,
+        "reply_for ignores delay_secs; delayed responses need harness support"
+    );
```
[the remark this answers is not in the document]
mbjorkqvist: Agreed — switched to `assert_eq!` in 7fea0cd for the same reason; a scenario relying on response delays now fails loudly in all build modes until delay support is added.

src/xrc-test-proxy/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "xrc-test-proxy"
```
gregorydemay, line 2: Understanding question: why build our own proxy canister and not use the [one](https://github.com/dfinity/proxy-canister) from the SDK team that works with ICP CLI?

scripts/pocket-ic-tests
```diff
@@ -0,0 +1,59 @@
...
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+WASM_DIR="${REPO_ROOT}/target/wasm32-unknown-unknown/release"
+
+# Match the forex behaviour the expected test values were computed with.
+export DISABLE_FOREX_WEEKEND_CHECK="yes"
+export DISABLE_FOREX_TIMEZONE_OFFSET="yes"
+
+# Build the XRC wasm (with the forex feature flags above) and the proxy wasm.
+"${REPO_ROOT}/scripts/build-wasm"
```
gregorydemay, line 51: For the Pocket IC tests it's usually worthwhile to have 2 variations:

1. CI takes the wasm produced by the reproducible build (the artifact that's released) and tests it via Pocket IC
2. To speed up local development one can allow the wasm just being built by a pure cargo command.

src/xrc-tests/src/tests/basic_exchange_rates.rs
```diff
@@ -82,177 +80,153 @@ fn basic_exchange_rates() {
     .chain(mock_responses::forex::build_common_responses(now_seconds))
     .collect::<Vec<_>>();
 
-    let container = Container::builder()
-        .name("basic_exchange_rates")
-        .exchange_responses(responses)
-        .build();
+    let env = XrcTestEnv::setup(responses, now_seconds);
```
gregorydemay, line 83: nit: I'm not sure I would pass the mock as a setup constructor parameter. Typically integration test setup are expensive to build (they take time, in the order of seconds) so it makes sense to reuse them as much as possible (potentially have multiple mocked scenarios with a single setup instance)

src/xrc-tests/src/pocket.rs
```diff
@@ -0,0 +1,184 @@
...
+const IDLE_ROUNDS_TO_SETTLE: usize = 5;
+
+/// A running PocketIC instance with the XRC and the cycles-forwarding proxy installed.
+pub struct XrcTestEnv {
+    pic: PocketIc,
+    proxy: Principal,
+    /// Outcall URL -> canned response. The URL is produced by the XRC's own
+    /// `exchange.get_url(..)` / `forex.get_url(..)`, the same functions `mock_responses` uses,
+    /// so the match is an exact string comparison.
+    mocks: HashMap<String, ExchangeResponse>,
```
gregorydemay, line 45: There I would try reuse as much as possible the logic from `ic-pocket-canister-runtime` for the mocks, e.g. it does allow to mock different results for the same url (e.g. first time called respond this, 2nd time called respond that), you also want to be sure that ALL mocks are used, etc.

gregorydemay (commented):
Thanks a lot @mbjorkqvist for starting this! Some understanding question from my side but otherwise generally LGTM

src/xrc-test-proxy/Cargo.toml
mbjorkqvist: Fair question, and honestly neither option is clean:
- Pull in `dfinity/proxy-canister`: a separate published artifact, so we'd either vendor a prebuilt wasm (opaque, hand-pinned) or add a build/git dependency — and it's generic, so the harness would have to hand-encode the inner `get_exchange_rate` call and decode a nested result.
- Keep a ~30-line in-repo proxy: stays on the XRC's own `ic-cdk` (independent of the DEFI-2878 bump) and is typed to `get_exchange_rate` so the harness calls it directly — but it's one more thing we maintain.

Neither is ideal. For this spike I went with the in-repo proxy as the lower-friction choice. It's worth revisiting if/when we pick the migration back up — especially alongside adopting `ic-canister-runtime`'s `CyclesWalletRuntime` post-DEFI-2878 — but I'd rather not block the proof-of-concept on settling it.

scripts/pocket-ic-tests
mbjorkqvist: Agreed in principle: CI ideally exercises the reproducible-build wasm and local runs use a plain `cargo build`. Today both go through `scripts/build-wasm`; the reproducible Docker build only runs in `docker-build.yml`/`release.yml`. The clean fix is to have `scripts/pocket-ic-tests` honor a pre-set `XRC_WASM_PATH`. I'm leaving it out here, though — finishing the PocketIC migration is a sizable effort and not currently the top priority in this repo, so this stays a proof-of-concept and I'd rather not grow its footprint. Noting it as a follow-up for when we resume.

src/xrc-tests/src/tests/basic_exchange_rates.rs
mbjorkqvist: Worth flagging the wrinkle: the forex mocks are consumed during `setup` itself (the heartbeat's forex outcalls populate the store before any call), so they have to be present at construction — and each scenario also pins its own clock (`now_seconds`) and forex dataset, so a shared, reused instance isn't free either. Splitting "bring up instance" from "register mocks" is the right move once there are enough scenarios for the setup cost to bite. Since the broader migration is paused and only one scenario is ported, I'll leave the simple form and revisit when we pick it back up.

src/xrc-tests/src/pocket.rs
mbjorkqvist: The plan is to adopt `ic-pocket-canister-runtime` wholesale rather than reimplement its mock logic — but its 0.5 release pins `ic-cdk` 0.20 / `pocket-ic` 13.x, exactly the DEFI-2878 coupling this spike deliberately avoids. Both gaps you flag are real (ordered multi-response per URL; panic-on-unused-mock vs our flag-the-inverse), and they matter more for the caching/misbehaviour scenarios than for `basic_exchange_rates`. Given the migration isn't currently top priority, I'd rather not hand-roll these now; they're tracked as part of the consolidation the PR description flags, for when we resume and the ic-cdk bump lands.

mbjorkqvist (commented):
Thanks for the feedback, @gregorydemay! I'm leaning towards postponing this effort for the moment, and picking it up again once some other the other related cleanup work is done. I'll put the PR back into draft mode for the time being.

PR #343, opened 2026-06-19
fix: DEFI-2896: never cache or return an empty/zero post-filter crypto rate

mbjorkqvist:
## Purpose

Fixes a cache-poisoning issue: the Exchange Rate Canister could return a successful `Ok(rate = 0)` for a crypto pair to a non-privileged caller, served from cache, even though the same data correctly fails validation on a fresh request.

The root cause is a validation-boundary gap: a crypto/USDT intermediate whose post-filter rate set is empty (e.g. every source reported a zero) was inserted into the cache before validation, and the cache-only crypto/crypto path returned without re-validating. A later request for the same asset and timestamp was then served the cached empty-rates entry, whose median is 0.

## What this changes

The validation boundary is closed at every layer so an empty/zero rate can neither be produced, cached, nor returned:

- The fresh fetch rejects an empty post-filter rate with `NoRatesFound`.
- The cache refuses to store a rate with an empty rates vector.
- The cache-only crypto/crypto path now validates the composed result, mirroring the fresh path.
- The exchange parser rejects non-finite, non-positive and over-range values before scaling, so an invalid sample never enters the rate set even if an exchange emits a valid-shaped zero.

After the fix, requests for assets with no rates return an error consistently instead of an occasional cached zero. Privileged callers were already protected by the minimum-rate cache guard.

## Tests

- A regression test (first commit) shows the empty intermediate was being cached; it fails on the pre-fix code.
- A new end-to-end test asserts a second, cache-only request re-fetches and errors rather than returning `Ok(rate = 0)`.
- A parser test asserts a zero price is rejected.

The candid compatibility check passes — no public API change (errors continue to use existing variants).

### DIFF
--- a/src/xrc/src/api.rs
+++ b/src/xrc/src/api.rs
@@ -113,16 +113,27 @@ impl CallExchanges for CallExchangesImpl {
             return Err(CallExchangeError::NoRatesFound);
         }
 
+        let queried_exchange_rate = QueriedExchangeRate::new(
+            asset.clone(),
+            usdt_asset(),
+            timestamp,
+            &rates,
+            exchanges.len(),
+            rates.len(),
+            None,
+        );
+
+        // The raw rates may all be filtered out (e.g. every source reported a
+        // zero or otherwise invalid price), leaving an empty post-filter rate.
+        // Such a rate must never be cached or returned: its median is zero, so a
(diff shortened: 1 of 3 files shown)


gregorydemay (approved):
Thanks @mbjorkqvist !

PR #345, opened 2026-06-19
fix: DEFI-2896: validate composed rate on the cache-only crypto/fiat path

mbjorkqvist:
## Purpose

Follow-up to #343, which closed the empty/zero-rate validation boundary on the cache-only crypto/crypto path but left the identical asymmetry in the crypto/fiat path.

The fresh crypto/fiat path validates the composed crypto/USD rate, but the cache-only branch returned it directly with `Ok(...)` and no `validate()`. As a result a composed rate the system would reject as `InconsistentRatesReceived` on a fresh request could be served as a success once its inputs were cached — i.e. the same request returned different results depending on cache state.

## What this changes

- The cache-only crypto/fiat branch now validates the composed result, mirroring the fresh path, so a cache-only response can no longer bypass the validation boundary.
- Adds a test asserting the cache-only crypto/fiat path rejects an inconsistent composed rate (fails on pre-fix code).
- Adds a test for the cache-only crypto/crypto path, which #343 fixed but left untested.
- Renames `cache_only_request_never_returns_zero_rate_after_failed_validation` to reflect what it actually asserts (the empty intermediate is not cached, so the next request re-fetches).

## Severity

This is not the critical zero-rate cache poisoning #343 fixed — that path is closed at the source (empty rates are neither produced nor cached). This closes a lower-severity consistency gap: a rate deemed inconsistent on a fresh fetch could still be served from cache. No zero/garbage rate results, but fresh and cache-only requests should agree.

### DIFF
--- a/src/xrc/src/api.rs
+++ b/src/xrc/src/api.rs
@@ -719,13 +719,16 @@ async fn handle_crypto_base_fiat_quote_pair(
     }
     let forex_rate = forex_rate_result?;
 
+    // We have all of the necessary rates in the cache; return the result.
+    // Validate the composed result here too, mirroring the fresh path, so a
+    // cache-only response can never bypass the validation boundary.
     if num_rates_needed == 0 {
         let crypto_base_rate =
             maybe_crypto_base_rate.expect("Crypto base rate should be set here.");
         let stablecoin_rate = stablecoin::get_stablecoin_rate(&stablecoin_rates, &usd_asset())
             .map_err(ExchangeRateError::from)?;
         let crypto_usd_base_rate = crypto_base_rate * stablecoin_rate;
-        return Ok(crypto_usd_base_rate / forex_rate);
+        return (crypto_usd_base_rate / forex_rate).validate();
     }
 
     with_inflight_tracking(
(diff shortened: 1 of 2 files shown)


src/xrc/src/api.rs
```diff
@@ -719,13 +719,16 @@ async fn handle_crypto_base_fiat_quote_pair(
...
+    // Validate the composed result here too, mirroring the fresh path, so a
+    // cache-only response can never bypass the validation boundary.
     if num_rates_needed == 0 {
         let crypto_base_rate =
             maybe_crypto_base_rate.expect("Crypto base rate should be set here.");
         let stablecoin_rate = stablecoin::get_stablecoin_rate(&stablecoin_rates, &usd_asset())
             .map_err(ExchangeRateError::from)?;
         let crypto_usd_base_rate = crypto_base_rate * stablecoin_rate;
-        return Ok(crypto_usd_base_rate / forex_rate);
+        return (crypto_usd_base_rate / forex_rate).validate();
```
gregorydemay, line 731: To avoid this type of error I would use 2 different types (can be done in a follow-up PR):

1. UnvalidtedType
2. ValidatedType. Can only be constructed from UnvalidatedType by calling the validating function.
3. We only store ValidatedType

gregorydemay (approved):
Thanks @mbjorkqvist !

src/xrc/src/api.rs
mbjorkqvist: Good point, I'll work on that in a follow-up PR as suggested!

mbjorkqvist: This doesn't seem to be a trivial change, so I created [DEFI-2907](https://dfinity.atlassian.net/browse/DEFI-2907) to track this (spun off from DEFI-2896).

PR #347, opened 2026-06-23
feat: DEFI-2903: record empty exchange windows as no_data, not http_error

mbjorkqvist:
## Why

Coinbase's stablecoin `http_error` rate sits near 50% in production — but those aren't transport failures. Coinbase doesn't forward-fill, so a minute with no trade returns an empty `[]`; the outcall transform trapped on that, surfacing as `CallExchangeError::Http` → `Outcome::HttpError`. That conflated "no trades this minute" with genuine HTTP failures and polluted the dashboard. The forex path already makes this distinction (`CallForexError::Empty` → `Outcome::EmptyMap`); the exchange-rate path now does too.

## Change

The empty case is already a distinct extractor result — `ExtractError::Extract` ("parsed, no datapoint") vs `ExtractError::JsonDeserialize` (malformed). The transform now signals it through the success channel instead of trapping:

- `encode_response`/`decode_response` carry `Option<u64>`;
- the transform emits `None` for `ExtractError::Extract` (genuine parse failures still trap);
- `call_exchange_raw` maps `None` → a new `CallExchangeError::NoData`;
- `record_exchange_outcome` records it as a new `Outcome::NoData` (`no_data`).

## Scope & safety

- **Applies to every exchange and both call kinds.** The reclassification keys on `ExtractError::Extract`, which all 9 exchange extractors produce only for an empty candle collection, so any exchange's empty crypto *or* stablecoin window now records `no_data` (an empty crypto window was previously mislabeled `http_error` too). This is intended.
- **Purely outcome classification.** The source set and `supported_stablecoin_pairs()` are unchanged — **Coinbase is kept** (its no-forward-fill behaviour is the *safe* one during a depeg: it returns no data rather than a stale price). No change for Poloniex/CryptoCom.
- **No rate change.** An empty window contributed nothing to the median before (it errored out) and contributes nothing now (`no_data`), so rates and the e2e golden values are unaffected — verified by a full local e2e run (6/6 green).
- `http_error` now reflects only real transport failures; `no_data` is a per-`(exchange, kind)` liquidity signal.

Tests: `encode`/`decode` `Option` round-trip incl. `None`; `extract_rate` empty→`Extract` vs malformed→`JsonDeserialize`; `record_exchange_outcome` → `no_data`; transform empty `[]` → `None` and a populated candle → `Some`. 262 unit tests + 6 e2e pass.

Jira: https://dfinity.atlassian.net/browse/DEFI-2903

### DIFF
--- a/src/xrc/src/exchanges.rs
+++ b/src/xrc/src/exchanges.rs
@@ -1428,13 +1428,17 @@ mod test {
     }
 
     /// The function tests the ability of [Exchange] to encode a response body from the
-    /// exchange transform function, including the empty/no-data `None` case.
+    /// exchange transform function, including the empty/no-data `None` case. The
+    /// wire format is candid `(opt nat64,)`; the exact bytes are pinned for both
+    /// cases (and shown to round-trip).
     #[test]
     fn encode_response() {
-        let bytes = Exchange::encode_response(Some(100)).expect("should be able to encode value");
-        assert!(matches!(Exchange::decode_response(&bytes), Ok(Some(100))));
+        let some = Exchange::encode_response(Some(100)).expect("should be able to encode value");
+        assert_eq!(hex::encode(&some), "4449444c016e780100016400000000000000");
+        assert!(matches!(Exchange::decode_response(&some), Ok(Some(100))));
 
         let none = Exchange::encode_response(None).expect("should be able to encode no-data");
+        assert_eq!(hex::encode(&none), "4449444c016e78010000");
         assert!(matches!(Exchange::decode_response(&none), Ok(None)));
     }
 
@@ -1449,13 +1453,15 @@ mod test {
     }
 
     /// The function tests the ability of [Exchange] to decode a response body from the
-    /// exchange transform function, for both a present rate and the no-data case.
+    /// exchange transform function, for both a present rate and the no-data case,
+    /// from the pinned candid `(opt nat64,)` bytes.
     #[test]
     fn decode_response() {
-        let some = Exchange::encode_response(Some(100)).expect("should be able to encode value");
+        let some =
+            hex::decode("4449444c016e780100016400000000000000").expect("should be able to decode");
         assert!(matches!(Exchange::decode_response(&some), Ok(Some(rate)) if rate == 100));
 
-        let none = Exchange::encode_response(None).expect("should be able to encode no-data");
+        let none = hex::decode("4449444c016e78010000").expect("should be able to decode");
         assert!(matches!(Exchange::decode_response(&none), Ok(None)));
     }
 
--- a/src/xrc/src/exchanges.rs
+++ b/src/xrc/src/exchanges.rs
@@ -1427,18 +1427,19 @@ mod test {
         assert_eq!(hex_string, "4449444c0001780000000000000000");
     }
 
-    /// The function tests the ability of [Exchange] to encode a response body from the
-    /// exchange transform function, including the empty/no-data `None` case. The
-    /// wire format is candid `(opt nat64,)`; the exact bytes are pinned for both
-    /// cases (and shown to round-trip).
+    /// The function tests that [Exchange] encodes and decodes a response body
+    /// (the transform's `Option<u64>` payload) symmetrically, for both a present
+    /// rate and the empty/no-data `None` case. The exact byte layout is not
+    /// pinned: `encode_response`/`decode_response` are an ephemeral within-outcall
+    /// round-trip in a single canister build (the bytes are never persisted or
+    /// decoded by another version), so a round-trip is the only property that
+    /// matters.
     #[test]
-    fn encode_response() {
+    fn encode_decode_response_round_trips() {
         let some = Exchange::encode_response(Some(100)).expect("should be able to encode value");
-        assert_eq!(hex::encode(&some), "4449444c016e780100016400000000000000");
         assert!(matches!(Exchange::decode_response(&some), Ok(Some(100))));
 
         let none = Exchange::encode_response(None).expect("should be able to encode no-data");
-        assert_eq!(hex::encode(&none), "4449444c016e78010000");
         assert!(matches!(Exchange::decode_response(&none), Ok(None)));
     }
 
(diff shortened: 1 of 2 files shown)


src/xrc/src/exchanges.rs
```diff
@@ -1424,12 +1428,18 @@ mod test {
...
+    /// exchange transform function, including the empty/no-data `None` case. The
+    /// wire format is candid `(opt nat64,)`; the exact bytes are pinned for both
+    /// cases (and shown to round-trip).
     #[test]
     fn encode_response() {
-        let bytes = Exchange::encode_response(100).expect("should be able to encode value");
-        let hex_string = hex::encode(bytes);
-        assert_eq!(hex_string, "4449444c0001786400000000000000");
+        let some = Exchange::encode_response(Some(100)).expect("should be able to encode value");
+        assert_eq!(hex::encode(&some), "4449444c016e780100016400000000000000");
```
gregorydemay: understanding question: why do we care about the exact encoding taken by `Exchange::encode_response`?

src/xrc/src/exchanges.rs
```diff
@@ -1443,13 +1453,16 @@ mod test {
...
-        let bytes = hex::decode(hex_string).expect("should be able to decode");
-        let result = Exchange::decode_response(&bytes);
-        assert!(matches!(result, Ok(rate) if rate == 100));
+        let some =
+            hex::decode("4449444c016e780100016400000000000000").expect("should be able to decode");
+        assert!(matches!(Exchange::decode_response(&some), Ok(Some(rate)) if rate == 100));
+
+        let none = hex::decode("4449444c016e78010000").expect("should be able to decode");
+        assert!(matches!(Exchange::decode_response(&none), Ok(None)));
     }
```
gregorydemay: Doesn't it duplicate the `encode_response` test?

src/xrc/src/lib.rs
```diff
@@ -253,6 +253,13 @@ pub(crate) enum Outcome {
     ExtractedZero,
     #[strum(serialize = "no_rates_found")]
     NoRatesFound,
+    /// The upstream returned a well-formed response with no datapoint — an
+    /// empty candle window (e.g. a thinly-traded pair with no trade in the
+    /// queried minutes on an exchange that does not forward-fill). This is
+    /// expected market behaviour, not a failure, so it is kept out of
+    /// `http_error`; a rising rate here is a liquidity signal for that pair.
+    #[strum(serialize = "no_data")]
+    NoData,
```
gregorydemay, line 268: Understanding question: what's the difference with `NoRatesFound`? Maybe adding a light Rust doc on that variant would help.

src/xrc/src/lib.rs
```diff
@@ -1184,7 +1211,15 @@ pub fn transform_exchange_http_response(args: TransformArgs) -> HttpResponse {
...
-        Ok(rate) => rate,
+        Ok(rate) => Some(rate),
+        // The response parsed fine but carried no datapoint — i.e. the exchange
+        // returned an empty candle window (e.g. Coinbase, which does not
+        // forward-fill, when no trade occurred in the queried minutes). That is
+        // "no data this minute", not a transport or parse failure, so signal it
+        // as `None` instead of trapping; the caller records it as a distinct
+        // `no_data` outcome rather than `http_error`. Genuine parse failures
+        // (malformed JSON, unconvertible numbers) still trap.
+        Err(ExtractError::Extract(_)) => None,
```
gregorydemay, line 1230: The doc on `ExtractError::Extract(_)` states:
```
    /// The filter failed to extract from the JSON as the filter selects a value improperly.
```
which sounds like a parsing error.

Looking at the code, it seems the only way to construct that error variant is via `extract_rate`:
```rust
    let extracted_value = extract_fn(response).ok_or_else(|| ExtractError::extract(bytes))?;
```

So would this change also impact other exchanges?

src/xrc/src/lib.rs
```diff
@@ -2368,6 +2403,82 @@ mod test {
...
+            #[allow(deprecated)]
+            let body = {
+                use ic_cdk::api::management_canister::http_request::{HttpResponse, TransformArgs};
+                let args = TransformArgs {
+                    response: HttpResponse {
+                        status: candid::Nat::from(200u64),
+                        headers: vec![],
+                        // Coinbase returns `[]` for a minute with no trade (it
+                        // does not forward-fill).
+                        body: b"[]".to_vec(),
```
gregorydemay, line 2460: I'm not sure there is a test for it (looked at the callers of `ExtractError::JsonDeserialize`) but maybe it would be good to also have a test to assert the error in case of Json parsing error?

gregorydemay (commented):
Thanks @mbjorkqvist ! Some minor comments

src/xrc/src/exchanges.rs
mbjorkqvist: Good question - the byte-pin turns out to guard nothing here. `encode_response`/`decode_response` are an ephemeral within-outcall round-trip in a single canister build: the bytes are encoded in the transform and decoded immediately in `call_exchange_raw`, never persisted, cached (the cache holds decoded `QueriedExchangeRate`, not candid bytes), or decoded by another canister version. So there's no earlier-version-decode scenario for the exact bytes to catch. Dropped the hex and kept a `Some`/`None` round-trip in c5dd41c.

src/xrc/src/exchanges.rs
mbjorkqvist: Agreed — once the byte-pin is gone, the two tests are pure round-trips and redundant. Consolidated them into a single round-trip test (c5dd41c).

src/xrc/src/lib.rs
mbjorkqvist: `NoData` is a single exchange's well-formed-but-empty response. `NoRatesFound` maps from `CallExchangeError::NoRatesFound` - the aggregate "nothing found for the asset" produced by the api-level rate collection (and the stablecoin invert-zero case) - which, unlike `NoData`, is never emitted per-exchange through `call_exchange`/`record_exchange_outcome`. Added a contrasting note to the `NoData` doc (696e41a).

src/xrc/src/lib.rs
mbjorkqvist: Yes, intentionally - it applies to every exchange and to both crypto and stablecoin calls (an empty crypto window was also mislabeled `http_error`). I audited all 9 extractors: each is `.first()/.data.first().map(...)`, so `None` only ever means "empty collection"; a malformed body fails earlier as `JsonDeserialize`, and `Extract` is constructed only in `extract_rate` (forex uses `RateNotFound`). You're right the `Extract` doc was misleading (legacy jq wording) - rewrote it, broadened the transform comment to state the cross-exchange scope (696e41a), and added a test pinning empty→`Extract` vs malformed→`JsonDeserialize` (c5dd41c).

src/xrc/src/lib.rs
mbjorkqvist: Added `extract_rate_distinguishes_empty_from_malformed` (c5dd41c): empty `[]` → `ExtractError::Extract` (→ no_data), malformed → `ExtractError::JsonDeserialize` (still traps → http_error), valid candle → `Ok`. Asserted at the extractor level since checking the transform's `ic0.trap` directly in a host test is messy.

gregorydemay (approved):

PR #341, opened 2026-06-16
fix: DEFI-2890: drop CryptoCom USDT-USDC stablecoin source

mbjorkqvist:
Drop CryptoCom as a USDT-USDC stablecoin source, since Crypto.com has completely delisted Tether (USDT) for European Economic Area (EEA) users to comply with the European Union's Markets in Crypto-Assets (MiCA) regulation.

PR #348, opened 2026-06-25
feat: DEFI-2904: freshness gate for forward-filled stablecoin sources (Poloniex)

mbjorkqvist:
## Why

Poloniex forward-fills its thin `USDT-USDC` market: when no trade occurs in a closed minute it still returns a candle carrying a stale price, so the fetch *succeeds* and the stale value silently enters the USDC median — harmless near the peg, dangerous during a depeg. DEFI-2903's empty-window → `no_data` reclassification can't catch this, because Poloniex never returns an empty `[]` window.

## What

A **stablecoin-only freshness gate**: a forward-filling source contributes a rate only when the candle reflects a real trade, otherwise `no_data`.

- The rate transform context now carries an `is_stablecoin` flag (`encode_rate_context`/`decode_rate_context`), so the gate applies to stablecoin calls only — crypto extraction is unchanged (crypto forward-fill is out of scope; see DEFI-2888).
- `Poloniex::extract_stablecoin_rate` treats a zero trade count (candle field 8) as no datapoint, which flows into the existing `no_data` plumbing from DEFI-2903 (`Option<u64>` → `CallExchangeError::NoData` → `Outcome::NoData`). No new outcome, metric, or API — the candle already carries the trade count.
- Other exchanges use the default (`= extract_rate`); the gate generalises to any future forward-filler by overriding the one method.

Result: Poloniex contributes a fresh price when it actually traded and `no_data` otherwise — the same honest behaviour Coinbase now has — so a stale forward-filled price never enters the median. Self-healing if liquidity returns.

Part 2 of the ticket (a sustained-`no_data` / "source effectively silent" observability signal) is alerting work and is left as a follow-up.

Jira: https://dfinity.atlassian.net/browse/DEFI-2904

## Source-count safety (no risk of running out of sources)

The gate only narrows the **USDC** pool, and only the two thin sources in it:

- **USDC — 8 sources:** six are the deeply liquid `USDC-USDT` markets (KuCoin, OKX, GateIo, MEXC, Bitget, Digifinex — ~100% of minutes traded, never gated), plus the two thin `USDT-USDC` sources, Coinbase and Poloniex. Even if both Coinbase (empty → `no_data`) and Poloniex (gated → `no_data`) drop out in a given minute, **6 liquid USDC rates always remain.**
- **USDS — 5 sources** (OKX, GateIo, MEXC, Bitget, Digifinex): Poloniex doesn't serve USDS, so the gate doesn't touch it at all.

The hard floor is `MIN_NUM_STABLECOIN_RATES = 2` *symbols* (USDC + USDS), each needing ≥1 exchange rate. With 6 and 5 real sources respectively, that floor holds with wide margin, so the gate cannot cause a "too few stablecoin rates" failure. And Poloniex's gated contribution was a **stale forward-filled price, not a genuine datapoint** — dropping it to `no_data` removes a fake source, not a real one, so the count of *real* sources goes up, not down.

### DIFF
--- a/src/xrc/src/exchanges.rs
+++ b/src/xrc/src/exchanges.rs
@@ -846,6 +846,25 @@ impl IsExchange for Poloniex {
         })
     }
 
+    /// Freshness gate for the stablecoin path. Poloniex forward-fills: when no
+    /// trade occurred in the queried (closed) minute it still returns a candle,
+    /// carrying the stale price. Field 8 of the candle is the trade count, so a
+    /// zero there means the price is a carried-over stale value. Treat that as
+    /// "no datapoint" ([ExtractError::Extract], which the transform maps to
+    /// no_data) instead of feeding a stale price into the stablecoin median.
+    /// A real trade (count > 0) contributes its price as usual.
+    fn extract_stablecoin_rate(&self, bytes: &[u8]) -> Result<u64, ExtractError> {
+        extract_rate(bytes, |response: PoloniexResponse| {
+            response.first().and_then(|kline| {
+                if kline.8 == 0 {
+                    None
+                } else {
+                    Some(ExtractedValue::Str(kline.2.clone()))
+                }
+            })
+        })
+    }
+
     fn listing_url(&self) -> &str {
         "https://api.poloniex.com/markets"
     }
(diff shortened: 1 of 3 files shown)


src/xrc/src/exchanges.rs
```diff
@@ -810,6 +846,25 @@ impl IsExchange for Poloniex {
...
+        extract_rate(bytes, |response: PoloniexResponse| {
+            response.first().and_then(|kline| {
+                if kline.8 == 0 {
+                    None
+                } else {
+                    Some(ExtractedValue::Str(kline.2.clone()))
+                }
+            })
+        })
+    }
```
gregorydemay, line 866: Understanding question: why do this only for Poloniex? For exchanges where we could somehow know whether some trading occur in that time window, we should discard any given rate if no trading occurred, no?

mbjorkqvist: Good question — the principle (drop a rate when no trade occurred) is right, but applying it to **every** source would currently cause `NO_RATES`, because the binding constraint is **per-symbol source redundancy**, not the principle itself.

I simulated a gate-all rule against the last ~7 days of 1-minute candles. A minute "fails" for a symbol when *no* source had a real trade in it; if a symbol then produces no rate, `stablecoin_rates.len()` falls below `MIN_NUM_STABLECOIN_RATES = 2` and the crypto/USD request fails.

**USDS** — only OKX/GateIo/MEXC/Bitget/Digifinex list `USDS-USDT`:

| source | market | minutes with a real trade (of ~10,200) |
|---|---|---|
| MEXC | USDS-USDT | 10,001 |
| Digifinex | USDS-USDT | 4,559 |
| Bitget | USDS-USDT | 1,469 |
| OKX | USDS-USDT | 91 |
| GateIo | USDS-USDT | 49 |

→ **158 minutes (1.55%) with no rate, longest outage 26 minutes.**

**USDC** — all 8 sources:

| source | market | minutes with a real trade (of ~10,200) |
|---|---|---|
| OKX | USDC-USDT | 10,091 |
| Digifinex | USDC-USDT | 10,081 |
| MEXC | USDC-USDT | 10,080 |
| Bitget | USDC-USDT | 10,176 |
| KuCoin | USDC-USDT | 9,436 |
| GateIo | USDC-USDT | 7,189 |
| Coinbase | USDT-USDC | 1,666 |
| Poloniex | USDT-USDC | 798 |

→ **0 minutes with no rate.**

USDS real liquidity is essentially MEXC-only (99.5%) while the other four forward-fill; stripping their stale candles to `no_data` drops USDS to zero sources in those minutes → only USDC remains → `TooFewRates`. USDC is immune: it has 6 liquid `USDC-USDT` sources at ~100%, which is exactly why gating Poloniex (a USDC source) is safe.

So the realistic options are:

1. **Gate all USDC sources** (both orientations, i.e., USDC-USDT and USDT-USDC) — safe (0 failures in the sim), but needs per-exchange volume/trade-count parsing for 6 more candle formats while those liquid sources essentially never forward-fill, so little marginal benefit for the added surface.
2. **Gate only Poloniex** (this PR) — the one known forward-filler, in a redundant pool; USDS is untouched (not listed there). Safe.
3. **Don't gate** (close the PR) — keeps Poloniex's stale forward-filled price in the USDC median; fine near the peg, risky during a depeg.
4. **Gate Poloniex now + generalize via monitoring** — add other forward-fillers only after the DEFI-2891 divergence / DEFI-2904 sustained-`no_data` signals flag them and we've confirmed the symbol's pool can absorb it.
5. **Gate USDS too, but make a missing USDS non-fatal** — the only way to safely gate USDS, but it changes the `MIN_NUM_STABLECOIN_RATES` semantics (USDS exists precisely to avoid USDC-only reliance), so it is a larger, separate design decision.

Gating crypto sources has the same hazard but worse — thin/long-tail assets (e.g. OGY) are single-sourced — which is DEFI-2888's scope, out of this PR.

How about making an upgrade including only [PR#341](https://github.com/dfinity/exchange-rate-canister/pull/341) and [PR#347](https://github.com/dfinity/exchange-rate-canister/pull/347) (these should continue improving the signal-to-noise ratio in the dashboards and alerts), monitoring the situation, investigating other options, and potentially revisiting this PR after that? For now, I'd move the PR back to draft.

PR #352, opened 2026-07-03
feat: DEFI-2916: add USDC as a privileged asset

mbjorkqvist:
## Purpose

The XRC throttles its HTTP outcalls with a self-imposed rate limiter. Privileged asset pairs bypass this limiter for current-rate requests (`timestamp=None`), so the exchange rates consumers rely on most stay available even under load. Until now the privileged crypto assets were BTC, ETH, ICP, and USDT.

USDC was missing from that set, which meant USDC-against-fiat and USDC/USDT requests were subject to the limiter and could be rejected with `RateLimited` — despite USDC being just as important to consumers. This PR adds USDC as a privileged asset so those pairs bypass the rate limiter for current-rate requests, matching the treatment of the other privileged cryptos.

## Change

- Add `USDC` to `PRIVILEGED_CRYPTO_ASSETS`.
- Extend the `should_identify_privileged_assets` unit test to cover USDC with full parity to the existing privileged cryptos: privileged when paired with a fiat currency or USDT (both directions), and not privileged when paired with another non-USDT crypto.

[DEFI-2916](https://dfinity.atlassian.net/browse/DEFI-2916)

[DEFI-2916]: https://dfinity.atlassian.net/browse/DEFI-2916?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/xrc/src/lib.rs
+++ b/src/xrc/src/lib.rs
@@ -88,7 +88,7 @@ const PRIVILEGED_CANISTER_IDS: [Principal; 3] = [
     Principal::from_slice(&[0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01, 0x01, 0x01]),
 ];
 
-const PRIVILEGED_CRYPTO_ASSETS: [&str; 4] = [BTC, ETH, ICP, USDT];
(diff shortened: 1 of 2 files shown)


gregorydemay (approved):
Thanks @mbjorkqvist !

